PR #40, opened 2018-09-30
add aliases to help command

aresko:
#32


Cmdr/BuiltInCommands/help.lua
```diff
@@ -14,7 +14,7 @@ return {
 	Run = function (context, commandName)
 		if commandName then
 			local command = context.Cmdr.Registry:GetCommand(commandName)
-			context:Reply(("Command: %s"):format(command.Name), Color3.fromRGB(230, 126, 34))
+			context:Reply(("Command: %s %s"):format(command.Name, command.Aliases and table.getn(command.Aliases) > 0 and ("(Aliases: %s)"):format(table.concat(command.Aliases, ", ")) or ""), Color3.fromRGB(230, 126, 34))
```
evaera: - Do not use deprecated function `table.getn`
- Add another separate reply call for Aliases, they should be on their own line

PR #45, opened 2018-10-09
Add a client-side clear command

### LINKED ISSUE #44: Add a clear command
A client-side command that would use the `DefaultUtil` group.
When run, it would simply remove all lines besides the input line from the console screen.

TripleFrequency:
Closes #44 

There is one limitation to this implementation, in that when you run the command you are left with a "Window cleared." text at the top of the GUI.

To get around this, a way to actually return nothing from a command may be necessary.

### DIFF
--- a/Cmdr/BuiltInCommands/clear.lua
+++ b/Cmdr/BuiltInCommands/clear.lua
@@ -22,6 +22,6 @@ return {
             end
             Frame.Size = Frame.Size - UDim2.new(0, 0, 0, -shrinkHeight)
         end
-		return nil
+		return "Window cleared."
 	end
 }
--- a/Cmdr/BuiltInCommands/clear.lua
+++ b/Cmdr/BuiltInCommands/clear.lua
@@ -12,14 +12,11 @@ return {
 		local frame = gui:WaitForChild("Frame")
 
 		if gui and frame then
-			local shrinkHeight = 0
 			for _, child in pairs(frame:GetChildren()) do
 				if child.Name == "Line" and child:IsA("TextLabel") then
-					shrinkHeight = shrinkHeight + child.Size.Y.Offset
 					child:Destroy()
 				end
 			end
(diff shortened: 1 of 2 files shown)


evaera:
Returning an empty string explicitly (`""`) will not add a new line to the console.

Cmdr/BuiltInCommands/clear.lua
```diff
@@ -0,0 +1,27 @@
...
+
+    Run = function()
+        local Player = Players.LocalPlayer
+        local Gui = Player:WaitForChild("PlayerGui"):WaitForChild("Cmdr")
+        local Frame = Gui:WaitForChild("Frame")
+
+        if Gui and Frame then
+            local shrinkHeight = 0
+            for _, child in pairs(Frame:GetChildren()) do
+                if (child.Name == "Line" and child:IsA("TextLabel")) then
```
evaera: Unnecessary parenthesis

Cmdr/BuiltInCommands/clear.lua
```diff
@@ -0,0 +1,27 @@
...
+	Name = "clear";
+	Aliases = {};
+	Description = "Clear all lines above the entry line of the Cmdr window.";
+	Group = "DefaultUtil";
+	Args = {};
+
+    Run = function()
+        local Player = Players.LocalPlayer
+        local Gui = Player:WaitForChild("PlayerGui"):WaitForChild("Cmdr")
+        local Frame = Gui:WaitForChild("Frame")
```
evaera: Block scoped variables should be camelCase

TripleFrequency:
I ran a small test after https://github.com/evaera/Cmdr/pull/45/commits/fa5153e9b53899e8dbc114e39343a3d6aefea682. There seems to be an issue related to having a blank string as a response, where the Cmdr window increases its size again. I will look into it when I get the chance, to verify if it is strictly related to the clear command's implementation.

PR #47, opened 2018-10-10
Add Registry Validation

TripleFrequency:
This implementation throws errors on the client if a registered command has both a `Data` and `Run` section declared.
It also errors if a registered type's name is `nil`, if it isn't a string, if it contains anything other than letters, or doesn't start with a lowercase letter.


TripleFrequency:
You'll have to update the base branch @evaera, I accidentally left it on the default `master`.

evaera:
Type names may contain numbers

PR #63, opened 2019-02-07
Ban command

ImmortalSlayer:
The following request is a created ban command that ban a user for a certain amount of time (using the new duration type) as well as a given reason.


LoganDark:
Most games have their own ban system. Integrators of Cmdr would most likely create one of their own.

Taking a quick glimpse of your code, I don't think it fits the theme of the other default commands.

AnonymousDeveloper13:
I honestly disagree and feel like a default system should be in place but you could modify it if you want to work with your own ban system

evaera:
If there was a ban system, it wouldn't be modifiable. The best you could do is just not register it and make your own. Cmdr's built in commands should never be modified manually.

LoganDark:
@evaera Maybe a ban system could serve as an example, but not a default feature of Cmdr.

PR #84, opened 2019-08-07
Fix multi-line text clipping.

Luke-Gnar:
Fixes multi-line text being clipped through everything in the console due to incorrect Y offset size in the textlabel.


evaera:
Please fix build before review.

Luke-Gnar:
I don't know how luacheck works. I'm unsure why the build is failing.

PR #105, opened 2020-03-06
Add utility method that takes server command implementations from a t…

FableRBX:
…able of functions in place of separate ModuleScripts.

An example of how this would be used:

`Cmdr:RegisterCommandsInContainerUsingTable(ServerScriptService.Commands, ServerCommands.Commands)`

_ServerCommands.lua_
```
local GameService

local ServerCommands = {}

function ServerCommands:Init()
    GameService = self.Services.GameService
end

function ServerCommands:OnCommandInvoked(context)
    print(string.format("%s invoked command %s!", context.Executor.Name, context.Name))
    return true
end

ServerCommands.Commands = {
    AddItem = function(context, toPlayer, itemType, itemId)
        local res = ServerCommands:OnCommandInvoked(context)
        if res then
            return GameService:AddItemCmd(toPlayer, itemType, itemId)
        else
            error(context.Executure.Name .. " is not allowed to run this command!")
        end
    end,

    AddResource = function(context, toPlayer, resourceType, resourceAmt)
        local res = ServerCommands:OnCommandInvoked(context)
        if res then
            return GameService:AddResourceCmd(toPlayer, resourceType, resourceAmt)
        else
            error(context.Executure.Name .. " is not allowed to run this command!")
        end
    end
}

return ServerCommands
```

### DIFF
--- a/Cmdr/Shared/Registry.lua
+++ b/Cmdr/Shared/Registry.lua
@@ -182,6 +182,24 @@ function Registry:RegisterCommandsIn (container, filter)
 	end
 end
 
+--- A helper method that registers all commands inside a specific container and sources the server implementations from a given table.
+function Registry:RegisterCommandsInContainerUsingTable (container, serverCmdTable)
+    for _, commandScript in pairs(container:GetChildren()) do
+        if commandScript:IsA("ModuleScript") then
+            local commandObject = require(commandScript)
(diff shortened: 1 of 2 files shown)


evaera:
Thanks for the PR, but I'm going to close this because I don't think it's good practice.

Issue #121, opened 2020-04-28
Lists


evaera:
- `list [something]`: 
  - `players`
- `pick [n] [list]`: Get a value from a list
- `len [list]`: Length of list
- `random [n]`: Get a random value between 1..n

Get a random player:
`pick ${run random ${len list players}} ${list players}`

Maybe: `pick !random [list]` as a built-in to pick from a list

TheNexusAvenger:
For my use case (Innovation Security's use case), trainers run commands on random players for specific rounds in the game (ex: `:bring random,random,random` to bring 3 random players, oh say to be snipers). Moving to Cmdr currently means the functionality doesn't exist. It would be helpful to have "random" being valid for the `player` and `players` type.

Issue #89, opened 2019-10-17
Make *, . work for all types


evaera:
If a type is listable, and it has an autocomplete function, then entering `*` should be interpreted as the list of every entry returned from the autocomplete function when there is no text for that argument.

Using a `.` in place of an argument value should be considered the "default value". It's only usable if the type definition has a function called `Default`. `Default` is called with the command context and it should return the value which the commands will get. For the `player` and `players` type, the Default will return the executing player. 

`**` should be interpreted as everything from `*`, minus the default value `.`. This is akin to the "others" shorthand that currently exists in the `players` type. Not sure on this one though. Other options could be `!`, `*!`, `*.`, or `*-.` (all minus default value).

After this, the "all", "others", and "me" shorthands will be removed from the `players` type.

evaera:
`?` for random

PR #122, opened 2020-04-28
Add Random Player Selector

TheNexusAvenger:
# Purpose
The purpose of this pull request is to add the ability to select random players with the `player` and `players` datatype.

# Justification
Innovation Security (managed by non-developers) runs security trainings for the group and performs events. Some of these events rely on picking random players (between 1 and n), such as for snipers to stop everyone else from reaching the end of a map. The current syntax used as part of the older admin system (Nexus Admin) uses `random` as the selector, such as `:bring random,random,random` to bring 3 random players.

# Implementation Overview
The change requested involves adding both ~`random` and~ `?` to the `player` type as selectors for selecting random players, as well as being able to select a specific amount of random players using ~`random[n]` and `?[n]`~ `?n` (as mentioned in https://github.com/evaera/Cmdr/issues/121). This should not close the issue since it doesn't provide a robust solution for randomness/lists and is specific to players.

Edit: `random` and `random[n]` were removed and `?[n]` was changed to `?n` as requested from private messages.

### DIFF
--- a/Cmdr/BuiltInTypes/Player.lua
+++ b/Cmdr/BuiltInTypes/Player.lua
@@ -30,7 +30,7 @@ local function ShorthandMultiple (text, executor)
 		if maxSize and maxSize > 0 then
 			local players = {}
 			local remainingPlayers = Players:GetPlayers()
-			for i = 1,math.min(maxSize, #remainingPlayers) do
+			for i = 1, math.min(maxSize, #remainingPlayers) do
 				table.insert(players,table.remove(remainingPlayers, math.random(1, #remainingPlayers)))
 			end
 			
--- a/Cmdr/BuiltInTypes/Player.lua
+++ b/Cmdr/BuiltInTypes/Player.lua
@@ -4,6 +4,13 @@ local Players = game:GetService("Players")
 local function ShorthandSingle (text, executor)
 	if text == "." or text == "me" then
 		return {executor}
+	elseif text == "random" or text == "?" then
+		local players = Players:GetPlayers()
+		if #players <= 1 then
+			return players
+		else
+			return {players[math.random(1,#players)]}
+		end
 	end
 end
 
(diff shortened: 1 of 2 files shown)


Cmdr/BuiltInTypes/Player.lua
```diff
@@ -19,6 +26,17 @@ local function ShorthandMultiple (text, executor)
 			end
 		end
 		return Others
+	elseif string.match(text, "%?(%d+)") then
+		local maxSize = tonumber(string.match(text, "%?(%d+)"))
+		if maxSize and maxSize > 0 then
+			local players = {}
+			local remainingPlayers = Players:GetPlayers()
+			for i = 1,math.min(maxSize, #remainingPlayers) do
+				table.insert(players,table.remove(remainingPlayers, math.random(1, #remainingPlayers)))
```
evaera, line 35:
```suggestion
				table.insert(players, table.remove(remainingPlayers, math.random(1, #remainingPlayers)))
```

Cmdr/BuiltInTypes/Player.lua
```diff
@@ -19,6 +26,17 @@ local function ShorthandMultiple (text, executor)
 			end
 		end
 		return Others
+	elseif string.match(text, "%?(%d+)") then
+		local maxSize = tonumber(string.match(text, "%?(%d+)"))
+		if maxSize and maxSize > 0 then
+			local players = {}
+			local remainingPlayers = Players:GetPlayers()
+			for i = 1,math.min(maxSize, #remainingPlayers) do
```
evaera, line 34:
```suggestion
			for i = 1, math.min(maxSize, #remainingPlayers) do
```

Cmdr/BuiltInTypes/Player.lua
```diff
@@ -19,6 +26,17 @@ local function ShorthandMultiple (text, executor)
 			end
 		end
 		return Others
+	elseif string.match(text, "%?(%d+)") then
+		local maxSize = tonumber(string.match(text, "%?(%d+)"))
```
evaera, line 30:
```suggestion
		return Others
	end
	
	local randomMatch = text:match("%?(%d+)")
	if randomMatch then
		local maxSize = tonumber(randomMatch)
```

Cmdr/BuiltInTypes/Player.lua
```diff
@@ -4,6 +4,13 @@ local Players = game:GetService("Players")
 local function ShorthandSingle (text, executor)
 	if text == "." or text == "me" then
 		return {executor}
+	elseif text == "?" then
+		local players = Players:GetPlayers()
+		if #players <= 1 then
+			return players
+		else
+			return {players[math.random(1, #players)]}
+		end
```
evaera, line 13:
```suggestion
		return {players[math.random(1, #players)]}
```

I think the performance gain is not worth the added complexity here

evaera:
Responded with a review. Commit the suggestions if they make sense or leave comments if anything is confusing

TheNexusAvenger:
I initially had the special casing for the `<= 1` case because I originally didn't know `math.random(1,1)` worked. The less than the case was to handle 0 because I instinctively thought "what if there are 0 players"... although on second thought, I would be concerned if that condition were to happen. I will commit the suggestions.

evaera:
> I initially had the special casing for the `<= 1` case because I originally didn't know `math.random(1,1)` worked. The less than the case was to handle 0 because I instinctively thought "what if there are 0 players"... although on second thought, I would be concerned if that condition were to happen. I will commit the suggestions.

That is an interesting thought. It's probably not possible right now but is something worth considering when implementing #89 for the more general case.

Issue #124, opened 2020-05-06
The cmdr input field is selectable using controller when GUI is not visible


robinwassen:
Cmdr is interfering with the selection cursor when playing with a controller.

**Affected versions:** 1.5.0, 1.4.0

**Steps to reproduce:**
- Play with a controller
- Press the back button to toggle the GUI selection 

**Expected results:** You should not be able to select the input field

**Actual results:** The cmdr input field is selectable

**Prosposed solution:**
- Add a selection group to the cmdr frame

![image](https://user-images.githubusercontent.com/302151/81140353-4f1f2280-8f69-11ea-8e83-4d05aa6e6a8d.png)

evaera:
We can fix this by setting [https://developer.roblox.com/en-us/api-reference/property/GuiObject/Selectable] to false in https://github.com/evaera/Cmdr/blob/master/Cmdr/CreateGui.lua

robinwassen:
@evaera I will get back with a PR in 15 minutes to fix this :)

PR #126, opened 2020-05-06
Fixed so TextBox is not selectable using controller #124

robinwassen:
See #124 

I took the liberty to remove the auto-generated notice. Judging by the files currently in version control I assume that was some legacy method of creating this file.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 # Next
 
 - Add random selector for player type (#122)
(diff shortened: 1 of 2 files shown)


evaera:
Yep! That's true. The comment is old.

Please update CHANGELOG.md under the Next title and reference this issue (#126)

robinwassen:
Nice! I have added a commit to update CHANGELOG.md now.

PR #128, opened 2020-05-27
Update Hooks.md

TsjipTsjip:
Note: This is a very minor change. I realise this and I understand if this gets rejected. I myself believe it would be a good contribution, hence the PR.

I ran into issues with my Client AfterRun hooks (unexpectedly) running after I interrupted the command execution in a Server BeforeRun hook. This PR changes a sentence that, to me, is ambiguous. While it's literally a single word to add, I believe it would prevent further confusion around this.


evaera:
I think this change makes the documentation inaccurate because whether or not the command will be interrupted depends on where that hook is running. I updated the PR to be more accurate with how Cmdr works internally

PR #132, opened 2020-05-29
Before run hook security

TsjipTsjip:
This is once again pretty rudimentary. Going to find out wether or not I can do this in a more efficient way or not before I'll mark this as ready for review.

Should the warning and error message be configurable? (Is there a way to make the string return appear more as an error, i.e. red text?)


TsjipTsjip:
Ready for review.

@evaera Do we need to change anything about this PR's behavior? I'm still looking for a way to make the error message red with a string return from the BeforeRun hooks section, but didn't have any luck.

evaera:
Hook output doesn't support coloration, just like return output from commands don't. Just a limitation of the system right now

evaera:
I think we probably should also block client commands from running. Just allow any command in the `DefaultUtil` group to go through, that should prevent output spam. I think that should be a good solution. 

I also think that we should only display that in-studio warning once, not for every command. 

And make sure to add a line to CHANGELOG.md

PR #133, opened 2020-05-29
Fix uninitialized variable Script Analysis warning

FableRBX:
Util.SplitString has an initialization line where 4 variables are assigned to an expression with two values. This causes a linter warning in Script Analysis in Studio - "Assigning 2 values to 4 variables initializes extra variables with nil; add 'nil' to value list to silence".


FableRBX:
Any chance of getting this merged?

evaera:
Somehow missed this!

PR #138, opened 2020-08-01
Add HasImplementation function to CommandContext

enescglyn:
I made a BeforeRun and AfterRun hook on client, it sends the info to remote function(to validate things on server) then returns the response, the thing is while client commands get validated and logged 1 times on the server, Server commands get 2 times because both the client sends info and server has the hooks.

This will allow me(and others) to check if the command is client sided then sends the info so it doesnt get double validated and logged.

### DIFF
--- a/Cmdr/Shared/Command.lua
+++ b/Cmdr/Shared/Command.lua
@@ -18,7 +18,14 @@ function Command.new (options)
 		Object = options.CommandObject; -- The command object (definition)
 		Group = options.CommandObject.Group; -- The group this command is in
 		State = {}; -- A table which will hold any custom command state information
-		HasClientRun = options.CommandObject.ClientRun and true or false; -- Checks if the command has client run function.
+		HasImplementation = function() --Returns true if the command has an implementation on the caller's machine.
+			if RunService:IsClient() then
+				return options.CommandObject.ClientRun and true or false
+			elseif RunService:IsServer() then
+				--Assuming the server implementation of the command is the Run function of CommandObject.
+				return options.CommandObject.Run and true or false
+			end
+		end
 		Aliases = options.CommandObject.Aliases;
 		Alias = options.Alias; -- The command name that was used
 		Description = options.CommandObject.Description;
--- a/Cmdr/Shared/Command.lua
+++ b/Cmdr/Shared/Command.lua
@@ -217,10 +217,10 @@ end
 --- Returns true if the command has an implementation on the caller's machine.
 function Command:HasImplementation()
 	if RunService:IsClient() then
-		return options.CommandObject.ClientRun and true or false
+		return self.Object.ClientRun and true or false
 	elseif RunService:IsServer() then
 		--Assuming the server implementation of the command is the Run function of CommandObject.
-		return options.CommandObject.Run and true or false
+		return self.Object.Run and true or false
 	end
 end
 
--- a/Cmdr/Shared/Command.lua
+++ b/Cmdr/Shared/Command.lua
@@ -18,6 +18,7 @@ function Command.new (options)
 		Object = options.CommandObject; -- The command object (definition)
 		Group = options.CommandObject.Group; -- The group this command is in
 		State = {}; -- A table which will hold any custom command state information
+		HasClientRun = options.CommandObject.ClientRun and true or false; -- Checks if the command has client run function.
 		Aliases = options.CommandObject.Aliases;
 		Alias = options.Alias; -- The command name that was used
 		Description = options.CommandObject.Description;
--- a/Cmdr/Shared/Command.lua
+++ b/Cmdr/Shared/Command.lua
@@ -18,14 +18,6 @@ function Command.new (options)
 		Object = options.CommandObject; -- The command object (definition)
 		Group = options.CommandObject.Group; -- The group this command is in
 		State = {}; -- A table which will hold any custom command state information
-		HasImplementation = function() --Returns true if the command has an implementation on the caller's machine.
-			if RunService:IsClient() then
-				return options.CommandObject.ClientRun and true or false
-			elseif RunService:IsServer() then
-				--Assuming the server implementation of the command is the Run function of CommandObject.
-				return options.CommandObject.Run and true or false
-			end
-		end
 		Aliases = options.CommandObject.Aliases;
 		Alias = options.Alias; -- The command name that was used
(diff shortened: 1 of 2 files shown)


evaera:
Can you provide more info on your use case for this?

enescglyn:
So let me tell you the thing:
-- My hooks needs to access the serverstorage to check if the player can run the command or not
With the cmdr when you register hooks using RegisterHooksIn function it gets registered both in client and server, when the client runs a command, the hook will error because it doesn't have access to things in the serverstorage.

I tried registering hooks in the server script that requires Cmdr, it worked but its only for commands that has Server run thingy, that is due to the execution order Cmdr has.
hcHn3EuGbY

Because of this, I made a remotefunction. The client will have a before run hook and it will return the thing what the remote function returns. It worked but the issue is the commands that has server run function get checked 2 times because of the hook on client and server. This was OK for me until i made a afterrun hook. Server commands would get logged 2 times. So I made the HasClientRun property, now what my clients before run hook would do is: Check if command context has client run, if yes then fire remote function, if it has no client run then do nothing. This worked for me and this is why i am making a pr for this

evaera:
So in addition to `ClientRun`, command objects can also have `Run`. You would need to check if either of these functions are present in the CommandObject.

Rename `HasClientRun` to `RunsOnClient`, as I think this name captures what's going on a little better.

evaera:
This presents another issue though: the `Run` method is automatically injected into the command object on the server. So if we check that this property will be wrong when checked from the server.

evaera:
I think it would be best to change this property to a function, called `HasImplementation`. It should check if there is a `ClientRun` or `Run` function present and return. 

This way, it's accurate based on where you're calling it from

evaera:
You can use this description for the function:


```
Returns `true` if the command has an implementation on this machine. For example, this function will return `false` from the client if you call it on a command that only has a server-side implementation.

Note that commands can potentially run on both the client and the server, so what this property returns on the server is not related to what it returns on the client, and vice versa. Likewise, receiving a return value of `true` on the client does not mean that the command won't run on the server, because Cmdr commands can run a first part on the client and a second part on the server.

This function only answers one question: if you run the command, does it run any code as a result of that on this machine?
```

Cmdr/Shared/Command.lua
```diff
@@ -18,6 +18,14 @@ function Command.new (options)
...
 		Group = options.CommandObject.Group; -- The group this command is in
 		State = {}; -- A table which will hold any custom command state information
+		HasImplementation = function() --Returns true if the command has an implementation on the caller's machine.
+			if RunService:IsClient() then
+				return options.CommandObject.ClientRun and true or false
+			elseif RunService:IsServer() then
+				--Assuming the server implementation of the command is the Run function of CommandObject.
+				return options.CommandObject.Run and true or false
+			end
+		end
```
evaera, line 28: This should go on the Command prototype. Creating a new function every time isn't great for memory

docs/api/CommandContext.md
```diff
@@ -22,6 +22,11 @@ docs:
     - name: State
       desc: A blank table that can be used to store user-defined information about this command's current execution. This could potentially be used with hooks to add information to this table which your command or other hooks could consume.
       type: table
+    - name: HasImplementation
+      desc: Returns `true` if the command has an implementation on the caller's machine. For example, this function will return `false` from the client if you call it on a command that only has a server-side implementation.
+      Note that commands can potentially run on both the client and the server, so what this property returns on the server is not related to what it returns on the client, and vice versa. Likewise, receiving a return value of `true` on the client does not mean that the command won't run on the server, because Cmdr commands can run a first part on the client and a second part on the server.
+      This function only answers one question if you run the command; does it run any code as a result of that on this machine?
+      type: function
```
evaera, line 29: You need to use the yaml multiline syntax
Example here: https://raw.githubusercontent.com/evaera/Cmdr/master/docs/api/Registry.md

image

Cmdr/Shared/Command.lua
```diff
@@ -214,4 +214,14 @@ function Command:GetStore(...)
 	return self.Dispatcher.Cmdr.Registry:GetStore(...)
 end
 
+--- Returns true if the command has an implementation on the caller's machine.
+function Command:HasImplementation()
+	if RunService:IsClient() then
+		return options.CommandObject.ClientRun and true or false
+	elseif RunService:IsServer() then
+		--Assuming the server implementation of the command is the Run function of CommandObject.
+		return options.CommandObject.Run and true or false
```
evaera, line 223: Where is `options` coming from?

docs/api/CommandContext.md
```diff
@@ -22,6 +22,12 @@ docs:
     - name: State
       desc: A blank table that can be used to store user-defined information about this command's current execution. This could potentially be used with hooks to add information to this table which your command or other hooks could consume.
       type: table
+    - name: HasImplementation
+      desc: |
+       Returns `true` if the command has an implementation on the caller's machine. For example, this function will return `false` from the client if you call it on a command that only has a server-side implementation.
+       Note that commands can potentially run on both the client and the server, so what this property returns on the server is not related to what it returns on the client, and vice versa. Likewise, receiving a return value of `true` on the client does not mean that the command won't run on the server, because Cmdr commands can run a first part on the client and a second part on the server.
+       This function only answers one question if you run the command; does it run any code as a result of that on this machine?
+      type: function
```
evaera, line 30: This is incorrect, please look at the other functions to learn how to write docs for them. You shouldn't put a function in the properties section

enescglyn: Ok, let me see.

Cmdr/Shared/Command.lua
```diff
@@ -214,4 +214,14 @@ function Command:GetStore(...)
...
 end
 
+--- Returns true if the command has an implementation on the caller's machine.
+function Command:HasImplementation()
+	if RunService:IsClient() then
+		return self.Object.ClientRun and true or false
+	elseif RunService:IsServer() then
+		--Assuming the server implementation of the command is the Run function of CommandObject.
+		return self.Object.Run and true or false
+	end
```
evaera, line 224: `Run` can also occur on client commands, actually.

```suggestion
	return ((RunService:IsClient() and self.Object.ClientRun) or self.Object.Run) and true or false
```

evaera:
Approved, going to test before merging

enescglyn:
Yay!

Issue #107, opened 2020-03-09
Math commands


evaera:
`math <operation> <value> <value>`

aliases / * + - %

and rand

PR #145, opened 2020-10-08
Add math utility commands

AnyThony:
#107


evaera:
Operation should be its own type. Would be nice to see built in aliases with AutoExec that provide things like `* 2 2`

evaera:
Since `^` is already taken by the history substitution, you can use `**` (which use used by other languages) for the exponentiation operation

evaera:
Change the operator type to be like this:

```lua
return function(registry)
	registry:RegisterType("operator", registry.Cmdr.Util.MakeEnumType("Operator", {
		{
			Name = "+";
			Perform = function (a, b)
				return a + b
			end
		};
		{
			Name = "*";
			Perform = function (a, b)
				return a * b;
			end
		};
                -- etc
	}))
end
```

Then in the command you can just do `operation.Perform(a, b)` and get rid of that giant if statement

evaera:
Closing as stale

PR #151, opened 2020-11-02
Fix GUI replicating twice

### LINKED ISSUE #125: Cmdr ScreenGui is added twice to PlayerGUI
This is probably a timing issue, so it might be hard to reproduce. But in our game the Cmdr ScreenGui is consistently added twice to the PlayerGUI in Studio (has not validated this in production).

**I assume that it this is caused by the following flow:**

1. Cmdr ScreenGui is added to StarterGui by the server
2. The client bootstrap script runs and require CmdrClient that causes the Cmdr ScreenGui to be copied to the PlayerGui
3. The Roblox default behaviour copies the CmdrClient from StarterGui to PlayerGui

**Affected versions:** 1.4.0, 1.5.0

**Screenshot of the issue:**

image

**The LocalScript we use to start Cmdr on the client (located in StarterPlayerScripts):**

```lua
local Players = game:GetService("Players")

local TGSMisc = require(workspace.Lib.TGSMisc)

if not TGSMisc.IsDeveloper(Players.LocalPlayer) then
	return
end

local Cmdr = require(ReplicatedStorage:WaitForChild("CmdrClient"))

Cmdr:SetActivationKeys({ Enum.KeyCode.BackSlash })
```

**Script used on the server:**

```lua

local Cmdr 		= require(workspace.Lib.External.Cmdr)
local TGSMisc 	= require(workspace.Lib.TGSMisc)


Cmdr.Registry:RegisterHook("BeforeRun", function(context)
	if not TGSMisc.IsDeveloper(context.Executor.UserId) then
		return "You don't have permission to run this command"
	end
end)

Cmdr:RegisterDefaultCommands()
Cmdr.Registry:RegisterCommandsIn(workspace.Lib.Commands)
```

y3th:
Fixes #125. Checks if the character has already been loaded before cloning in the GUI.

### DIFF
--- a/Cmdr/CmdrClient/init.lua
+++ b/Cmdr/CmdrClient/init.lua
@@ -39,7 +39,7 @@ local Cmdr do
 	Cmdr.Dispatcher = require(Shared.Dispatcher)(Cmdr)
 end
 
-if StarterGui:WaitForChild("Cmdr") and wait() and Player:WaitForChild("PlayerGui"):FindFirstChild("Cmdr") == nil then
+if Player.Character ~= nil and StarterGui:WaitForChild("Cmdr") and Player:WaitForChild("PlayerGui"):FindFirstChild("Cmdr") == nil then
 	StarterGui.Cmdr:Clone().Parent = Player.PlayerGui
 end
 


y3th:
Just realized this doesn't work when the character is never loaded (i.e. custom character controller).

PR #164, opened 2021-05-06
Organize the help command

Code-Sync:
Categorize the commands displayed by the groups pulled from each command definition for easier readability.

### DIFF
--- a/Cmdr/BuiltInCommands/help.lua
+++ b/Cmdr/BuiltInCommands/help.lua
@@ -1,4 +1,4 @@
-local ARGUMENT_SHORTHANDS = ([[
+local ARGUMENT_SHORTHANDS = [[
 Argument Shorthands
 -------------------			
 .   Me/Self
@@ -6,7 +6,7 @@ Argument Shorthands
 **  Others 
 ?   Random
 ?N  List of N random values			
-]])
+]]
 
 return {
 	Name = "help";
--- a/Cmdr/BuiltInCommands/help.lua
+++ b/Cmdr/BuiltInCommands/help.lua
@@ -1,3 +1,13 @@
+local ARGUMENT_SHORTHANDS = ([[
+Argument Shorthands
+-------------------			
+.   Me/Self
+*   All/Everyone 
+**  Others 
+?   Random
+?N  List of N random values			
+]])
+
 return {
 	Name = "help";
 	Description = "Displays a list of all commands, or inspects one command.";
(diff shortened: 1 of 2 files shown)


Cmdr/BuiltInCommands/help.lua
```diff
@@ -29,9 +29,27 @@ return {
...
+			
+			context:Reply([[
+Notes to consider!
+------------------
+. = Me/Self
+* = All/Everyone
+** = Others
+? = Random
+?N = List of N random values	
+]])
```
evaera, line 42: We should move this string to a constant at the top of the file so the indentation isn't as much of a problem.

I'd also separate them with tab characters instead of equal signs.

Replace "Notes to consider!" with "Argument Shorthands"

Cmdr/BuiltInCommands/help.lua
```diff
@@ -29,9 +29,27 @@ return {
...
+			for _, groupName in pairs(groups) do
+				context:Reply(([[
+%s
+------------------
+]]):format(groupName))
+				for _, cmd in pairs(commands) do
+					if cmd.Group == groupName then
+						context:Reply(("%s - %s"):format(cmd.Name, cmd.Description))
+					end
+				end
```
evaera, line 52: Having a nested for loop here is unnecessary and means we have to loop over *every* command for each group.

Instead, we should get the list of commands, sort the table by their group, and then loop over the list once. When the group changes from the previous group, insert a heading line with the group name. We also don't need to use the multi line literal syntax here.

Cmdr/Shared/Registry.lua
```diff
@@ -215,6 +215,20 @@ end
...
+	for _, group in pairs(Registry:GetCommands()) do
+		if not table.find(groups, group.Group) then
+			table.insert(groups, group.Group)
+		end
+	end
+	table.sort(groups)
+	
+	return groups
+end
+
```
evaera, line 231: I think we should remove this function as it isn't providing any new information. We can derive this from the command list if it is needed. With the updated implementation I described above, we don't need a list of groups anyways.

evaera (changes requested):

Cmdr/BuiltInCommands/help.lua
```diff
@@ -29,9 +39,17 @@ return {
...
 			local commands = context.Cmdr.Registry:GetCommands()
-
-			for _, cmd in pairs(commands) do
-				context:Reply(("%s - %s"):format(cmd.Name, cmd.Description))
+			table.sort(commands, function(a, b)
+				return a.Group < b.Group
+			end)
+			context:Reply(ARGUMENT_SHORTHANDS)
+			local lastGroup
+			for _, command in pairs(commands) do
```
evaera, line 47: We should use `ipairs` here because we're iterating over a list.

Cmdr/BuiltInCommands/help.lua
```diff
@@ -29,9 +39,17 @@ return {
 			end
 		else
 			local commands = context.Cmdr.Registry:GetCommands()
-
-			for _, cmd in pairs(commands) do
-				context:Reply(("%s - %s"):format(cmd.Name, cmd.Description))
+			table.sort(commands, function(a, b)
+				return a.Group < b.Group
+			end)
+			context:Reply(ARGUMENT_SHORTHANDS)
```
evaera, line 45: This line feels like it's right in the middle of the group logic. It should be moved up above the `local commands =` line, and add a space below it.

evaera (changes requested):

Cmdr/BuiltInCommands/help.lua
```diff
@@ -1,3 +1,13 @@
+local ARGUMENT_SHORTHANDS = ([[
+Argument Shorthands
+-------------------			
+.   Me/Self
+*   All/Everyone 
+**  Others 
+?   Random
+?N  List of N random values			
+]])
```
evaera, line 9: Excess parenthesis surrounding multiline string literal

```suggestion
local ARGUMENT_SHORTHANDS = [[
Argument Shorthands
-------------------			
.   Me/Self
*   All/Everyone 
**  Others 
?   Random
?N  List of N random values			
]]
```

evaera (changes requested):

evaera (approved):

PR #168, opened 2021-05-16
Replace TeleportPartyAsync with TeleportAsync

wilyt1:
TeleportPartyAsync can only teleport to a place in the same universe. I replaced it with the new TeleportAsync, which can teleport to places in a different universe.


evaera:
Will this keep the same group of players together on the other end?

wilyt1:
Unfortunately, I could not find anything regarding teleporting players to the same server instance on the DevHub.

Hexcede:
> Will this keep the same group of players together on the other end?

No, it would need to conditionally fall back to `TeleportAsync` when the target place is not in the same universe.

evaera:
Let's revert this PR back to the original commit (just simply replacing the function) and merge.

wilyt1:
I have reverted the pull request to the original commit.

PR #171, opened 2021-05-28
Update Dispatcher.lua to be Roblox compliant

railworks2:
Third party links (such as this) are not acceptable under 11. Other activities of the Roblox Community rules.

I wouldn't want Cmdr to be taken down over such an oversight

### DIFF
--- a/Cmdr/Shared/Dispatcher.lua
+++ b/Cmdr/Shared/Dispatcher.lua
@@ -144,7 +144,7 @@ function Dispatcher:RunHooks(hookName, commandContext, ...)
 
 		if RunService:IsStudio() then
 			if displayedBeforeRunHookWarning == false then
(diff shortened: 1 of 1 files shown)


evaera:
This error message is only displayed in Studio, so I believe it's already compliant with the rules.

railworks2:
I'm not aware of any expectations related to this being in studio. 

Given it's code present in a production module, I would expect to meet all of the Roblox Terms.

evaera:
This isn't going to cause moderation action against anyone. Having a link is very useful for beginners and I don't think there's any tangible value from removing it here. If it bothers you, feel free to remove it from your own copy.

enescglyn:
you are most craziest guy i have ever seen in your roblox profile's description you literally link your website and break roblox rule 11

enescglyn:
im surprised that roblox put you onto their post approval team

PR #174, opened 2021-06-20
Prevent quadratic backtracking in TrimString function

Validark:
The pattern "^%s*(.-)%s*$" doesn't do well for strings that have a large number of whitespace characters between non-whitespace characters.

Inspired by: https://stackstatus.net/post/147710624694/outage-postmortem-july-20-2016
Test bench: https://gist.github.com/Validark/c86a3e92dde76d26e6893d796dfea89a

### DIFF
--- a/Cmdr/Shared/Util.lua
+++ b/Cmdr/Shared/Util.lua
@@ -219,8 +219,9 @@ function Util.MashExcessArguments(arguments, max)
 end
 
 --- Trims whitespace from both sides of a string.
-function Util.TrimString(s)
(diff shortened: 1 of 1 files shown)


evaera:
Thanks for the PR! If we're going to make a change to fix the quadratic backtracking, it should be more thoroughly documented and more explicit. Particularly string.match with empty parenthesis is obscure compared to string.find. 

I'm not super concerned about this function because it's not run on untrusted input on the server, and I think having the simpler pattern might be worth keeping.

PR #176, opened 2021-07-01
Fix small spelling mistake

JPod32:
Change "this **is** a server directory" to "this **in** a server directory"


evaera:
This PR is changing an auto-generated file. This is the correct file to change: https://github.com/evaera/Cmdr/blob/master/docs/guide/Setup.md

PR #177, opened 2021-08-10
Improve run-lines Functionality

TheNexusAvenger:
While working with Innovation Security, we discovered a couple of things with respect to `run-lines` when using `init-edit` and `init-run`:
- By default, it doesn't appear to work at all since it uses `Dispatcher:Run(...)`, which only works on the client, but is called by the server.
- `run-lines` does not work with client commands, like `alias`.
- `run-lines` throws an error if the input text contains a percent character anywhere.

This pull request attempts to address those problems by making `run-lines` run on the client instead of the server, and escape percent characters before being used with `gsub`.

### DIFF
--- a/Cmdr/BuiltInCommands/Utility/runLines.lua
+++ b/Cmdr/BuiltInCommands/Utility/runLines.lua
@@ -12,12 +12,12 @@ return {
 		}
 	};
 
-	Run = function(context, text)
+	ClientRun = function(context, text)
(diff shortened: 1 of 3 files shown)


TheNexusAvenger:
The pull request had an additional change that wasn't intended for the pull request which now conflicts with master. I am going to re-do the pull request and update my fork so I can update a system I have that uses the fork.

PR #179, opened 2021-08-12
Change boolean to enum type

SlipperySpelunky:
This change makes a boolean primitive type into an enum and is parsed into a boolean value. This makes commands utilizing booleans in their arguments slightly quicker to use.

### DIFF
--- a/Cmdr/BuiltInTypes/Primitives.lua
+++ b/Cmdr/BuiltInTypes/Primitives.lua
@@ -39,28 +39,19 @@ local intType = {
 }
 
 local boolType do
-	local truthy = Util.MakeDictionary({"true", "t", "yes", "y", "on", "enable", "enabled", "1", "+"});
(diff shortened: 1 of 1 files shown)


evaera:
I don't understand how this makes it easier to use. Can you elaborate why you're making this change? 

The proposed change also does not have feature parity with the existing behavior.

SlipperySpelunky:
> I don't understand how this makes it easier to use. Can you elaborate why you're making this change? 
> 
> The proposed change also does not have feature parity with the existing behavior. 

It makes more sense for the possible options of a boolean to be displayed, as it's true/false. It looks like bad UX to validate this without showing what the user should be inputting before-hand.

evaera:
If all we want to change is adding an autocomplete menu, then the solution is to just add an `Autocomplete` function to the boolean type -- not to change it to an Enum type. It *isn't* an enum type right now - it accepts many different options which would be broken with the proposed change. I'm open to adding an autocomplete menu, but we don't need to change the existing code so drastically.

PR #198, opened 2021-11-15
Add nonNegativeInteger and positiveInteger types

### LINKED ISSUE #69: More number types
`natural`, which would be a subset of the current `integer` type, would allow numbers in the range `0..n`.
`positive`, which would be a subset of `natural`, would allow the range `1..n`

`byte`, which would allow the range `0..255`.
`digit`, which would allow the range `0..9`.

Quantify-RBLX:
Fixes #69 

nonNegative and positive were chosen over natural and positive due to dispute if zero is a natural number. Negative number types are not included due to that positive should be preferred.


evaera:
I'm curious why this was closed?

Quantify-RBLX:
I misattributed all the commits to my other account.

Quantify-RBLX:
#199

PR #210, opened 2022-02-21
Fix nil errors when using duration type

Famous5000:
Been having some "harmless" errors happening client-side whenever a Duration autocomplete command is typed in. {duration being the last argument}

This pull request aims to fix that. If you need to edit it to keep with your standards, then go ahead.


wilyt1:
Have you tried the changes from #159? The current release does not include changes from that pull request, so you may not have tested them. As well, I do not think the source should be edited to fix errors coming from a specific type.

Famous5000:
I have not tried them yet, this was mostly just to stop the error from being thrown due to duration being nil but judging by the code you wrote it seems like you fixed the main part of the problem.

Famous5000:
I'll leave this request up until an update to main has been pushed and the issue's been resolved. Contributors are free to close this without merging if the issue I'm trying to fix seems to have been fixed by that push.

PR #224, opened 2022-07-30
Fix issues with teleportation

DevAX1T:
Because of changes, "me" does not work anymore. I simply replaced "${me}" with ".".

this is also my first ever github pull request so please excuse my lack of knowledge


DevAX1T:
the "Update teleport.lua" commit actually shows the change you care about.
the "Fix issues with teleportation" commit is just a semicolon because i didn't know i could just describe the pull request here.

evaera:
Not the same thing. "me" is different from "${me}". That runs the `me` command, which returns the player's name.

PR #228, opened 2022-08-19
Output RichText

### LINKED ISSUE #157: Add RichText compatibility
Make all TextLabels and all TextBoxes so RichText tags can be used in responses and input

shane-sommer:
richtext enabled on output messages and changed "Line" to a textlabel since it wasn't being treated as a textbox

couldn't enable richtext on the entry line due to the way ui text rendering behaves
image

Closes #157

### DIFF
--- a/Cmdr/CmdrClient/CmdrInterface/Window.lua
+++ b/Cmdr/CmdrClient/CmdrInterface/Window.lua
@@ -59,8 +59,12 @@ function Window:UpdateWindowHeight()
 end
 
 --- Add a line to the command bar
-function Window:AddLine(text, color)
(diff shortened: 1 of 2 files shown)


shane-sommer:
Closes #157

evaera:
Not sure if I want to merge this. I don't particularly like RichText and I'm concerned about escaping RichText that is included in user input. If anything, it needs to be strictly opt-in per command

shane-sommer:
I have changed it to be opt-in like you said, replacing the "color" parameter in the AddLine method with an options table (includes backwards compatibility)

PR #232, opened 2022-09-03
Custom Window Height

Nidoxs:
When working with Cmdr in my projects, I sometimes needed the Cmdr UI to use more vertical screen real estate so I could see verbose messages that were returned from my Cmdr commands.

I've added this in so we can now size the window's height to a height that works best for our individual use cases.

**_Thanks._**


Nidoxs:
As a reminder, I did make a PR for this a few months ago but accidently ran my StyLua formatter on it and forgot to revert my changes.

This should now be fixed and hopefully won't intrude on your existing style used in this project.

PR #233, opened 2022-09-27
Improve run-lines Functionality

TheNexusAvenger:
Redo of #177

While working with Innovation Security, we discovered a couple of things with respect to `run-lines` when using `init-edit` and `init-run`:
- By default, it doesn't appear to work at all since it uses `Dispatcher:Run(...)`, which only works on the client, but is called by the server.
- `run-lines` does not work with client commands, like `alias`.
- `run-lines` throws an error if the input text contains a percent character anywhere.

This pull request attempts to address those problems by making `run-lines` run on the client instead of the server, and escape percent characters before being used with `gsub`.

### DIFF
--- a/Cmdr/BuiltInCommands/Utility/runLines.lua
+++ b/Cmdr/BuiltInCommands/Utility/runLines.lua
@@ -12,7 +12,7 @@ return {
 		}
 	};
 
-	Run = function(context, text)
+	ClientRun = function(context, text)
 		if #text == 0 then
 			return ""
 		end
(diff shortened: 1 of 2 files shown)


TheNexusAvenger:
Is there a reason why this was closed? The group this change was for has been relying on it without issue for a while now and we are stuck on using a fork of Cmdr because of it.

evaera:
Sorry, I was just trying to clean up some old issues. I think I might have misunderstood this PR when I glanced at it, it looks good to merge actually

TheNexusAvenger:
All good, thanks. Now we can use the main release instead of the fork.

PR #244, opened 2023-03-05
Mention Wally as a setup option in docs

seyaidev:
I didn't know that it was available on Wally, so it'd be helpful to mention for people already using it for their projects. Included under `server-dependencies` to emphasize it being installed under a server directory like ServerStorage


evaera:
Should ideally be in its own section

PR #245, opened 2023-03-05
Change var and varSet Commands to Not Yield on Require

### LINKED ISSUE #243: Change var and varSet commands to not yield on require
This has started causing problems with initialization recently, and seems to be the root cause of what was described in #241

wilyt1:
Closes #243


evaera (changes requested):
Can you change this so that if a command is run before these are set that the command yields until the initial operation is done? Right now it seems like it would be possible to miss initial commands.

PR #246, opened 2023-03-06
Fix Help Command Bug

### LINKED ISSUE #240: Commands without a group result in the help command erroring
As per the [API Reference](https://eryn.io/Cmdr/api/Registry.html#commanddefinition), group is optional.
Here's a tiny repro (in luau): https://github.com/sasial-dev/cmdr-repro

wilyt1:
Closes #240


sasial-dev (approved):
I've tested, and confirmed it works with my repro :)

PR #251, opened 2023-04-03
refactor(math): make arg order make sense!

sasial-dev:


sasial-dev:
Small little PR to improve the `math` command

evaera:
I don't think we need to make this change. It might make sense, but this is breaking change and could break user setups for no real reason

sasial-dev:
Fair enough - that's reasonable!

PR #254, opened 2023-04-07
Add Further Validation to Command Objects

wilyt1:


Cmdr/Shared/Registry.lua
```diff
@@ -134,6 +134,11 @@ end
 -- Handles replicating the definition to the client.
 function Registry:RegisterCommand (commandScript, commandServerScript, filter)
 	local commandObject = require(commandScript)
+	if typeof(commandObject) ~= "table" then
```
autonordev, line 137: Could use `assert` instead

autonordev (changes requested):

autonordev (approved):

PR #256, opened 2023-04-14
Add CTRL+C shortcut to clear current command

OttoHatt:
Common shortcut seen in bash, command prompt, etc. Not a proper behaviour but it's become a habit for me, and I think others. Useful for when you make too many typos and just want to start over.

### DIFF
--- a/Cmdr/CmdrClient/CmdrInterface/Window.lua
+++ b/Cmdr/CmdrClient/CmdrInterface/Window.lua
@@ -298,6 +298,17 @@ function Window:BeginInput(input, gameProcessed)
 			wait()
 			self:SetEntryText(self:GetEntryText())
 		end
+	elseif input.KeyCode == Enum.KeyCode.C and input:IsModifierKeyDown(Enum.ModifierKey.Ctrl) then
+		-- Shortcut to delete any entered text.
+		-- Common shortcut used in bash, command prompt, etc. to quickly restart typing a command.
+		-- I think techincally it kills whatever process is reading input, but...
+		if Entry.TextBox.SelectionStart == -1 then
+			-- Only delete if the user has no selected text.
+			-- Otherwise they'd CTRL+C to copy to clipboard and delete everything.
+			self:SetEntryText("")
+			-- I think that we want this.
+			self:ClearHistoryState()
+		end
 	else
 		self:ClearHistoryState()
 	end


autonordev:
1. What's wrong with <kbd>Ctrl</kbd> + <kbd>Backspace</kbd> (or <kbd>Ctrl</kbd> + <kbd>A</kbd> & <kbd>Backspace</bd>)

2. Why might clearing the history state be desirable, especially since you describe the feature as "clearing the current command"?

Also in Bash, cmd, etc <kbd>Ctrl</kbd> + <kbd>C</kbd> ends the process (i.e. the process asking for input); it doesn't clear the input line if you're typing in commands directly, which is how Cmdr works.

Cmdr/CmdrClient/CmdrInterface/Window.lua
```diff
@@ -298,6 +298,17 @@ function Window:BeginInput(input, gameProcessed)
...
 		end
+	elseif input.KeyCode == Enum.KeyCode.C and input:IsModifierKeyDown(Enum.ModifierKey.Ctrl) then
+		-- Shortcut to delete any entered text.
+		-- Common shortcut used in bash, command prompt, etc. to quickly restart typing a command.
+		-- I think techincally it kills whatever process is reading input, but...
+		if Entry.TextBox.SelectionStart == -1 then
+			-- Only delete if the user has no selected text.
+			-- Otherwise they'd CTRL+C to copy to clipboard and delete everything.
+			self:SetEntryText("")
+			-- I think that we want this.
```
autonordev, line 309: Could be better worded as `Clear autocomplete state`

Cmdr/CmdrClient/CmdrInterface/Window.lua
```diff
@@ -298,6 +298,17 @@ function Window:BeginInput(input, gameProcessed)
 			wait()
 			self:SetEntryText(self:GetEntryText())
 		end
+	elseif input.KeyCode == Enum.KeyCode.C and input:IsModifierKeyDown(Enum.ModifierKey.Ctrl) then
+		-- Shortcut to delete any entered text.
+		-- Common shortcut used in bash, command prompt, etc. to quickly restart typing a command.
+		-- I think techincally it kills whatever process is reading input, but...
```
autonordev, line 304: This comment and the one above it is unnecessary.

evaera:
I don't think it makes sense to implement this with Ctrl+C. This shortcut is already used for copying text, I don't think overloading that is necessary.

The shortcut that *actually* does this in bash is Ctrl+U. However, I'm not sure I'm totally convinced that users would expect to use this keybind in a Roblox GUI and have it work. Pressing Ctrl+a and then starting to type your new command is still the same number of key strokes and is far more standard and already works today

PR #261, opened 2023-05-15
Add ArgumentOperatorAliases to Player types

benhdev:
Allows you to use "me", "all", "others" and "random" as an argument shorthand for Player and PlayerId types. 

Doesn't prevent you from using the default shorthands.

image
image
image
image

### DIFF
--- a/Cmdr/BuiltInTypes/Player.lua
+++ b/Cmdr/BuiltInTypes/Player.lua
@@ -23,6 +23,13 @@ local playerType = {
 	Default = function(player)
 		return player.Name
 	end;
+
+	ArgumentOperatorAliases = {
+		me = ".";
+		all = "*";
+		others = "**";
+		random = "?";
+	};
 }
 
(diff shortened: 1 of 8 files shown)


Cmdr/BuiltInTypes/Player.lua
```diff
@@ -23,6 +23,13 @@ local playerType = {
 	Default = function(player)
 		return player.Name
 	end;
+
+	ArgumentOperatorAliases = {
```
autonordev, line 27: Should these (specifically all and others) be included within the singular Player type?

benhdev: Would make it easier for set up. Otherwise you'd have to edit the Util module, since MakeListableType just sets the ArgumentOperatorAliases on the listable version of the type.

When it is in singular context the all and others keywords arent used by default due to functionality in Argument:Transform

PR #263, opened 2023-05-24
Implement auto scroll to the bottom of the canvas on user text input

anyastrophic:
This makes the canvas auto scroll to the bottom on any text input.
On by default, call `Cmdr:CanvasAutoScrollToBottom()` with param `isEnabled` to set the behaviour

image


evaera:
I don't really think this should be off by default. I think it should just always be on

anyastrophic:
> I don't really think this should be off by default. I think it should just always be on

Alright! There you go. I think it should still be up to the user if they want this feature or not, though

anyastrophic:
@autonordev but the UI doesn't auto scroll on input atm... this is what makes it autoscroll, and also implements a toggle

evaera:
Please remove the new API surface. I don't think we need a way to toggle this.

wilyt1 (approved):

PR #267, opened 2023-06-14
Add “Tab→” Note on Autocompletion Entries

### LINKED ISSUE #70: Add "Tab→" note on autocompletion entries
Imagine this:

Type `h`. You see `help`, but you want `history`. Down arrow to select `history`, review the description, looks good. Right arrow... doesn't autocomplete, nothing happens. That's okay... clicking on the item also does nothing... Well heck, now I have to type a whole 5 letters. This is definitely not okay.

:)

wilyt1:
Closes #70

### DIFF
--- a/Cmdr/CreateGui.lua
+++ b/Cmdr/CreateGui.lua
@@ -20,14 +20,19 @@ return function()
 	Frame.AutomaticCanvasSize = Enum.AutomaticSize.Y
 	Frame.Parent = Cmdr
 
-	local Autocomplete = Instance.new("Frame")
+	local Autocomplete = Instance.new("ScrollingFrame")
 	Autocomplete.BackgroundColor3 = Color3.fromRGB(59, 59, 59)
 	Autocomplete.BackgroundTransparency = 0.5
 	Autocomplete.BorderSizePixel = 0
+	Autocomplete.CanvasSize = UDim2.new(0, 0, 0, 0)
 	Autocomplete.Name = "Autocomplete"
 	Autocomplete.Position = UDim2.new(0, 167, 0, 75)
+	Autocomplete.ScrollBarThickness = 6
+	Autocomplete.ScrollingDirection = Enum.ScrollingDirection.Y
+	Autocomplete.Selectable = false
 	Autocomplete.Size = UDim2.new(0, 200, 0, 200)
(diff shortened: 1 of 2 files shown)


autonordev:
This appears even while the tab option would be inappropriate or not do anything:

image
image

Cmdr/CreateGui.lua
```diff
@@ -179,20 +179,33 @@ return function()
...
 	Suggest.TextXAlignment = Enum.TextXAlignment.Left
 	Suggest.Parent = TextButton
 
+	local Footnote = Instance.new("TextLabel")
+	Footnote.BackgroundColor3 = Color3.fromRGB(255, 255, 255)
+	Footnote.BackgroundTransparency = 1
+	Footnote.Font = Enum.Font.Code
+	Footnote.Name = "Footnote"
+	Footnote.Size = UDim2.new(1, 0, 1, 0)
+	Footnote.Text = "Tab→"
```
autonordev, line 203: This is bikeshedding so feel free to ignore - but putting a space between Tab and the arrow feels better.

image
image

On the other hand, putting a space also feels wrong - just in a different way; I'm torn on this.

autonordev (changes requested):
image

There's an issue of overlap in larger options -- we could truncate the original text here (but afaik there's no other circumstance in which we truncate autocomplete options), or make the footnote somehow not appear if it would intrude on the option, or some other solution.

autonordev:
With the exception of the two comments I've raised, this all looks good to me and tests well in game, but I have a gut feeling that this change shouldn't be implemented. This is a high-impact and noticeable change and I'm unsure about the value that the footnote would provide, except for first-time users.

This is a bit of a philosophical issue I suppose but from my pov one does not simply use Cmdr with absolutely no knowledge (i.e. without being told by at least one other person what it is, how to activate it, what commands they can use, and by extension how to use the autocomplete). Are there other areas of the Cmdr user experience that are unclear (like activation keys); if so, are we going to add labels there too?

Although as part of my testing I noticed that for some reason the autocomplete cursor (highlight) is a TextButton with AutoButtonColor enabled, although clicking on it will just cause Cmdr to unfocus (and thus - unless set otherwise - close).

PR #268, opened 2023-06-15
Fix AutoComplete Menu Max Size

### LINKED ISSUE #118: AutoComplete menu does not have a clamped vertical max size
Currently, the Autocomplete menu does not have a max size on the vertical axis, causing it to flow off the screen in certain cases (an example may be having lots of tools which are available to be given).

There should be a maximum size for this menu, with the ability to "scroll" through the menu using the arrow keys when you reach the bottom of the visible menu.

wilyt1:
Closes #118


autonordev (approved):
Amazing!

PR #288, opened 2023-07-20
Fix default command group formatting

cmumme:


autonordev:
Thanks for the interest in contributing but the docs site is currently not being updated (i.e. changes made to mainline don't deploy to the site) and this change would be of minimal practical value.

PR #290, opened 2023-07-23
Fix `help [command]` Broken by Inline Arguments

### LINKED ISSUE #280: `help [command_name]` is broken by inline arguments
Using inline arguments will cause the `help [command_name]` command to break. More specifically, it will output an error when it tries to output information for the argument, as inline arguments are functions and not tables.

[This is the offending line](https://github.com/evaera/Cmdr/blob/master/Cmdr/BuiltInCommands/help.lua#L41):

```lua
for i, arg in ipairs(command.Args) do
	context:Reply(
		`#{i} {arg.Name}{if arg.Optional == true then "?" else ""}: {arg.Type} - {arg.Description}`
	)
end
```

As to the solution, I'm not sure what we should do here. We can't exactly inspect inline arguments because - by their very nature (being fed the specific context) - they are dynamic. Maybe we just skip over and output a generic message at the end that some arguments may vary?

wilyt1:
Closes #280

### DIFF
--- a/Cmdr/BuiltInCommands/help.lua
+++ b/Cmdr/BuiltInCommands/help.lua
@@ -38,7 +38,9 @@ return {
 			context:Reply(command.Description, Color3.fromRGB(230, 230, 230))
 			for i, arg in ipairs(command.Args) do
 				context:Reply(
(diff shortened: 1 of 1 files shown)


autonordev (approved):
untested but lgtm

PR #291, opened 2023-07-29
Add command-specific guard methods

### LINKED ISSUE #86: Command-specific guards
Similar to BeforeRun, except specific to each command. They would be functions specified in a `Guards` array in the command definition. Each function is called in sequence, receiving the context. If one returns a string, then that becomes the output and command is interrupted.

Use case: when you want to restrict specific commands behind some requirement in a reusable way. For example, some commands only work when the player has a character (or one that is alive). Instead of checking for that inside the context each time, you could just make it into a guard.

itsrenderman:
Closes #86 

- Guard methods are checked before hooks.

- When a guard method is called, the first parameter is the command context.

- The dispatcher also supports sending additional parameters as varargs (similar to hooks), though this is not used in standard command execution.

- If the guard method returns a non-nil value, the command will be interrupted and the value printed to the window.

**Example usage:**
```
return {
	-- <insert command object here with name, aliases, description, etc.>

	Guards = {
		function(self)
			if math.random(1,3) == 2 then
				return "You are unlucky!"
			end
		end;
		function(self)
			return if self.Executor.UserId ~= 92658764 then "You are not allowed to do this!" else nil
		end;
	};
}
```

### DIFF
--- a/Cmdr/Shared/Command.lua
+++ b/Cmdr/Shared/Command.lua
@@ -27,6 +27,7 @@ function Command.new (options)
 		Arguments = {}; -- A table which will hold ArgumentContexts for each argument
 		Data = options.Data; -- A special container for any additional data the command needs to collect from the client
 		Response = nil; -- Will be set at the very end when the command is run and a string is returned from the Run function.
+		Guards = options.Guards; -- A table of functions where the command will be interrupted if a string is returned
 	}
 
 	setmetatable(self, Command)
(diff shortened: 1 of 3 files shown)


wilyt1 (approved):
Tested!

itsrenderman:
> > Guard methods are checked before hooks
> 
> Out of curiosity can you explain why you decided to do it this way?

I interpreted the initial proposal with the idea that guards should have a higher potency on the outcome of command execution over hooks

autonordev:
Hey, thanks for the contribution. We've discussed this internally and think that hooks should run before guards, this is because hooks are more 'generic'. For instance, if a player lacks the permissions to run a command, that is more important than an argument needing to have a player whose character health is above 0.

Are you willing to make this adjustment yourself? If not, that's completely fine since I can do it.

itsrenderman:
Quickly changed this on phone, should be good now 👍

PR #311, opened 2023-09-03
Prepend Error and Warning Messages With [Cmdr]

### LINKED ISSUE #306: Prepend error and warning messages with [Cmdr]
This will make it more clear to developers and users reporting issues to developers that a warning or error is being emitted by Cmdr. 

While this does not break any API contracts, it is a change in behaviour, and some developers may rely on Cmdr's warning and error messages to look a certain way. For instance, in `pcall`s. Accordingly, this should be part of a minor change and should be signposted in the changelog entry / release notes.

wilyt1:
Closes #306

### DIFF
--- a/Cmdr/CmdrClient/init.lua
+++ b/Cmdr/CmdrClient/init.lua
@@ -7,7 +7,7 @@ local Util = require(Shared:WaitForChild("Util"))
 
 if RunService:IsClient() == false then
 	error(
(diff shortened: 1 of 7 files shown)


NidhoRoblox:
Why not a function that returns a warning/error with "[Cmdr] ", or "[Cmdr] ".. ?

PR #328, opened 2024-06-29
Add Relative Vector3 Type

### LINKED ISSUE #68: Relative Vector3 type
Like a `relative3` type that accepts `~` coordinates like Minecraft does. `to @ ~, ~500, ~` would teleport you 500 studs into the air.

Interpreting what these coordinates are relative _to_ would be at the script's discretion.

wilyt1:
Closes #68 with @itsrenderman's type definition, incorporating minor styling adjustments. Furthermore, updates the built-in teleport command to use the new relative Vector3 type.

**Declarations**:


Cmdr/BuiltInTypes/Vector.lua
```diff
@@ -0,0 +1,71 @@
...
+local vector2Type = Util.MakeSequenceType({
+	ValidateEach = validateVector,
+	TransformEach = tonumber,
+	Constructor = Vector2.new,
+	Length = 2,
+})
+
+local relativeVector3Type = {
+	Transform = function(text, executor)
+		local currentPosition = executor.Character:GetPivot().Position
```
ShadowDaughter: We should always ensure that the Character exists.

Cmdr/BuiltInTypes/Vector.luau
```diff
@@ -0,0 +1,71 @@
...
+	Length = 2,
+})
+
+local relativeVector3Type = {
+	Transform = function(text, executor)
+		local currentPosition = executor.Character:GetPivot().Position
+
+		return Util.Map(Util.SplitPrioritizedDelimeter(text, { ",", "%s" }), function(value, index)
+			if value:sub(1, 1) ~= "~" then
+				return tonumber(value)
```
ShadowDaughter, line 34: This would return nil if the given text is a string such as `~meow`.

ShadowDaughter (changes requested):

ShadowDaughter:
I also don't think calling this a strict "Relative Vector3" type is accurate. I believe calling it something like PositionVector3 would be more accurate

Cmdr/BuiltInCommands/Admin/teleport.lua
```diff
@@ -5,7 +5,7 @@ return {
 	Group = "DefaultAdmin",
 	AutoExec = {
 		'alias "bring|Brings a player or set of players to you." teleport $1{players|players|The players to bring} ${me}',
-		'alias "to|Teleports you to another player or location." teleport ${me} $1{player @ vector3|Destination|The player or location to teleport to}',
+		'alias "to|Teleports you to another player or location." teleport ${me} $1{player @ relativeVector3|Destination|The player or location to teleport to}',
```
ShadowDaughter: Since we renamed the type to `positionVector3` we need to update the definition too.

Cmdr/BuiltInCommands/Admin/teleport.lua
```diff
@@ -14,7 +14,7 @@ return {
 			Description = "The players to teleport",
 		},
 		{
-			Type = "player @ vector3",
+			Type = "player @ relativeVector3",
```
ShadowDaughter: Same as Line 8

Cmdr/BuiltInCommands/Admin/teleport.lua
```diff
@@ -14,7 +14,7 @@ return {
...
 			Name = "Destination",
 			Description = "The player to teleport to",
```
ShadowDaughter: Maybe update the description to elaborate that the value can be a position rather than just a player?

Cmdr/BuiltInTypes/Vector.lua
```diff
@@ -0,0 +1,76 @@
...
+})
+
+local vector2Type = Util.MakeSequenceType({
+	ValidateEach = validateVector,
+	TransformEach = tonumber,
+	Constructor = Vector2.new,
+	Length = 2,
+})
+
+local positionVector3Type = {
```
ShadowDaughter: As noted within the Discord, it might be worth seeing if MakeSequenceType can't do what is being done here already.

ShadowDaughter (changes requested):

Cmdr/BuiltInTypes/Vector.luau
ShadowDaughter: Using type annotation for an internal method here while the rest of Cmdr doesn't have any sort of type annotation is kind of weird. That said tho, the specified return type here is actually incorrect. It should be `(boolean, string?)` rather than just a boolean.

Cmdr/BuiltInTypes/Vector.luau
```diff
@@ -1,8 +1,10 @@
 local Util = require(script.Parent.Parent.Shared.Util)
```
ShadowDaughter, line 2: Possibly remove the whitespace? It's a nitpick but I don't see a reason to seperate these two variables

ShadowDaughter (changes requested):

ShadowDaughter (approved):
LGTM

PR #331, opened 2024-07-13
Refactor `rand` Command to Use Random Class

### LINKED ISSUE #329: Update `rand` to use Random class
The `rand` command currently uses `math.random` to generate numbers. Math.random is not actually random; it does it use a very good source of entropy by default, and the common approach is to set the seed (which is global across the realm) to the UNIX timestamp (which is also not _that_ good of a randomness source, and is not something we can expect to happen in modern games which are unlikely to use math.random in the first place).

This ticket proposes:
- the `rand` command creates a top-level variable with a `Random.new()` instance
- that Random instance is used in place of `math.random`

While simple, this does raise a question that I think need discussion before this is ready for work: **should we use integers or decimals?** IIRC with `math.random` it will use integers if the seed is an integer and decimals if the seed has a decimal point, but I think this is wrong. In-keeping with current behaviour would be to use integer, but the command will probably be more useful (e.g. for pairing with `pick` via meta-commands[^1]) with integers.

It's worth noting that even if we stick with decimals, there will be a change in behaviour, as developers will no longer be able to control the seed used (instead, it would be an internal source of entropy as with any other `Random.new()` calls). I can _maybe_ imagine some use cases for this, but I expect it would be pretty rare. I think a note in the changelog about this is fine, and the change would be suitable in a minor (feature) release.

[^1]: `pick` is typed to use a positive integer, which should error if given a decimal, so the `pick` call in `run echo My favourite fruit is ${pick ${rand 1 5} apples,pears,oranges,mangoes,bananas}` _should_ error.

wilyt1:
Fixes #329 by using `Random.new()` instance instead of `math.random()`. Also updates function parameter names to match command arguments and reduce ambiguity.

**Declarations**:

### DIFF
--- a/Cmdr/BuiltInCommands/Utility/rand.lua
+++ b/Cmdr/BuiltInCommands/Utility/rand.lua
@@ -1,3 +1,5 @@
+local generator = Random.new()
+
 return {
 	Name = "rand",
 	Aliases = {},
@@ -17,7 +19,11 @@ return {
 		},
 	},
 
-	ClientRun = function(_, min, max)
-		return tostring(max and math.random(min, max) or math.random(min))
+	ClientRun = function(_, firstNumber, secondNumber)
+		return tostring(
+			if secondNumber
+				then generator:NextInteger(firstNumber, secondNumber)
+				else generator:NextInteger(1, firstNumber)
+		)
 	end,
 }


autonordev (approved):
LGTM but please carefully test to make sure this works as expected (if you haven't already).

PR #332, opened 2024-08-04
Update Window.lua

lucasnielsen:
This simple addition fixes an issue that causes the keyboard to completely break on mobile devices. [Streamable link](https://streamable.com/sctoqj)

The script before would cause :CaptureFocus() everytime the TextBox would lose focus, creating an infinite loop of the keyboard opening and closing extremely quickly (as shown in the video link above).

My fix to this is adding a check if the user is on mobile or not. If they are on mobile, it will stop `Entry.TextBox:CaptureFocus()` from being ran.

**Declarations**:


wilyt1 (approved):
Tested!

PR #341, opened 2024-12-12
Move from .lua file extension to .luau #334

### LINKED ISSUE #334: Move from .lua file extensions to .luau
I can't imagine any potential breakage here, unless you're modifying the source (which we don't support).

GitHub now supports Luau, so nothing needs to be added to `.gitattributes`, and this also means that GitHub's syntax highlighting will show Luau features (whereas the current Luau highlighting doesn't).

On VSCode: older Lua extensions, including ones built for Roblox, will not understand `.luau` extensions, but this shouldn't be an issue for anyone using `luau-lsp`, and I assume `roblox-lsp` supports them.

> **This is not open to contributors, except for the maintainers. A pull request resolving this is not desired**; it would cause a lot of noise on the commit history because of how GitHub handles pull requests.

itsrenderman:
- Changes all `.lua` file extensions to `.luau`
- Closes #334 

**Declarations**:

- [X] I declare that this contribution was created in whole or in part by me.


itsrenderman:
Ah I understand.

> Was this PR motivated by struggles with GitHub file formatting (i.e. of Luau features)? If so, we could temporarily change the gitattributes to markup/highlight .lua files as Luau.

There were no motivations besides simply noticing files were still using the `.lua` extension instead of `.luau` while messing around with various things in the code.

I made the file name changes and noticed an issue was open with a matching description. I didn't notice the remark inside the issue stating that a pull request wasn't desired - my apologies.

> I do already have in mind how and when this change should be done, and there is no particular rush for it.

Any more information on this?

PR #342, opened 2024-12-27
Support player display names and merge Util.GetNames with transformInstanceSet

### LINKED ISSUE #335: Players type that supports display names
Inspired by [this DevForum post](https://devforum.roblox.com/t/cmdr-a-fully-extensible-and-type-safe-command-console-for-roblox-developers/182815/324?u=automationeer). Obviously we can't (it's not licenced) copy this code 1:1, nor should we (it's not particularly good), but it did give me the idea.

Perhaps we can create a prefix within `players` for this player by display name type. While `@` is only meant to be used before usernames, many places - including the official Creator Hub - use `@[display name]` (¬_¬) so I don't see an issue with us doing it.

The obvious caveat is that display names are not unique, so this type would need some way to discriminate between duplicates.

itsrenderman:
### Display Name Support

Related to #335 

This pull request adds support for Player.DisplayName in command argument completions.

Usernames are now displayed one of two ways:
- DisplayName (Username)
- Username

Only the player's username is displayed if the Player's `DisplayName` and `Name` are the same.

This functionality is implemented through the `Util.lua` module, eliminating the need to modify the various different built-in types (such as Player, PlayerId, Team).

This should also assist with compatibility of any custom types created by Cmdr users, assuming they are also using the utility functions available through Cmdr.

<details>

<summary><strong>Example showing players with a different DisplayName and Name</strong></summary>

Screenshot 2024-12-27 121841

</details>

<details>

<summary><strong>Example showing a player with the same DisplayName and Name</strong></summary>

Screenshot 2024-12-27 122021

</details>

---

### Merge Util.GetNames with transformInstanceSet

I made this change as a result of seeing both methods having a very similar functionality.

The only difference was `Util.GetNames` supported values not needing a `Name` property, defaulting to `tostring` when it didn't exist.

Adding this functionality to `transformInstanceSet` has no breaking effects on any internal code, and any users interacting with Cmdr's utiltiies should also not be affected, given `transformInstanceSet` is not accessible by default.

This change is essentially just the removal of unnecessary code repetition.

---

**Declarations**:


itsrenderman:
It should also be noted that players are only able to use alphanumeric characters in their display names, thus preventing any forms of trickery with the name completions.

image

wilyt1 (approved):

wilyt1:
It appears that this is not currently compatible with listable types. See #398 for details.

PR #352, opened 2025-06-01
Update Window.lua

VladimirLuciver1:
To make it easier for players to close the command window?

### DIFF
--- a/Cmdr/CmdrClient/CmdrInterface/Window.lua
+++ b/Cmdr/CmdrClient/CmdrInterface/Window.lua
@@ -248,6 +248,11 @@ function Window:BeginInput(input, gameProcessed)
 		return
 	end
 
+	if input.KeyCode == Enum.KeyCode.Escape then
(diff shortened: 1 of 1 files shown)


0x-Offset (changes requested):
this does not work, the text box stays focussed but i found a fix

RoWeb-Advertisers:
> ```lua
> if input.KeyCode == Enum.KeyCode.Escape then
> 	if self:IsVisible() then
> 		self:Hide()
> 		task.wait()
> 		Entry.TextBox:ReleaseFocus()
> 	end
> 	return
> end
> ```

escape is not a good key to set this as..

wilyt1:
Please apply the changes discussed above. Also, could you ensure the key is changeable?

wilyt1:
Closing this PR due to inactivity and unresolved merge conflicts. Feel free to open a new PR if you decide to update the implementation.

PR #376, opened 2026-07-10
add sorting to FuzzyFinder

DAMIANOO9:
add sorting to FuzzyFinder based on how early the key was found and value length; makes autocomplete way more reliable by prioritizing results which have the key at the start of the value

**Declarations**:


wilyt1 (approved):

PR #378, opened 2026-07-17
Implement command priority rules

### LINKED ISSUE #339: Handling name conflicts between a builtin and developer-specified command or type
> If two commands are registered with the same name, the last one registered replaces the original. But since we can't predict if users register default or custom commands last, it could lead to unintended overwriting of custom commands by default ones.

_Originally posted by @wilyt1 in https://github.com/evaera/Cmdr/issues/327#issuecomment-2221263008_

I think it's an accurate assumption that most people are registering built-in stuff first, but you know what they say about assumptions! Imo it risks introducing breaking changes, violating both semver and Cmdr's reputation for stability.

If there's a conflict between a built-in command or type and a developer-registered one, the developer one should take priority irrespective of which was registered later. For a conflict between two developer-registered commands, the one registered later should take priority (I believe this is the current behaviour.)

A warning should be emitted in either case so that way there's no surprises for the developer.

wilyt1:
Closes #339. Implements a command-override and registration-priority system. This handles collisions by prioritizing developer commands over default commands, and newer commands over older ones. When an overwrite occurs, the stale command's ModuleScript is destroyed to prevent replication leaks and avoid redundancy in ReplicatedRoot.

**Declarations**:

### DIFF
--- a/Cmdr/Shared/Registry.luau
+++ b/Cmdr/Shared/Registry.luau
@@ -155,6 +155,7 @@ local Registry = {
 		"Data",
 		"Group",
 		"Guards",
+		"_scriptInstance",
 	}),
 	CommandArgProps = Util.MakeDictionary({ "Name", "Type", "Description", "Optional", "Default" }),
 	Types = {},
@@ -277,20 +278,22 @@ Registry.RegisterHooksIn = Registry.RegisterTypesIn
 	Prefer using Registry:RegisterCommand for proper handling of client/server model.
 
 	@param commandObject CommandDefinition
+	@param isDefault boolean? -- True if this is a built-in Cmdr command
+	@return boolean -- Returns true if registration succeeded or false if aborted due to a developer override
 	@private
 	@within Registry
 ]=]
-function Registry:RegisterCommandObject(commandObject)
-	for key in pairs(commandObject) do
+function Registry:RegisterCommandObject(commandObject, isDefault: boolean?): boolean
+	for key in commandObject do
 		if self.CommandMethods[key] == nil then
 			error(`[Cmdr] Unknown key/method in command "{commandObject.Name or "unknown command"}": {key}`)
 		end
 	end
 
 	if commandObject.Args then
-		for i, arg in pairs(commandObject.Args) do
+		for i, arg in commandObject.Args do
 			if type(arg) == "table" then
-				for key in pairs(arg) do
+				for key in arg do
 					if self.CommandArgProps[key] == nil then
 						error(
 							`[Cmdr] Unknown property in command "{commandObject.Name or "unknown"}" argument #{i}: {key}`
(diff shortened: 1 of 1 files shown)


Cmdr/Shared/Registry.luau
```diff
@@ -277,20 +278,22 @@ Registry.RegisterHooksIn = Registry.RegisterTypesIn
...
-	for key in pairs(commandObject) do
+function Registry:RegisterCommandObject(commandObject, isDefault: boolean?): boolean
+	for key in commandObject do
 		if self.CommandMethods[key] == nil then
 			error(`[Cmdr] Unknown key/method in command "{commandObject.Name or "unknown command"}": {key}`)
 		end
 	end
 
 	if commandObject.Args then
-		for i, arg in pairs(commandObject.Args) do
```
ShadowDaughter, line 291: Writing a for loop without `pairs` works but is generally more confusing to read.

Cmdr/Shared/Registry.luau
```diff
@@ -277,20 +278,22 @@ Registry.RegisterHooksIn = Registry.RegisterTypesIn
...
+		for i, arg in commandObject.Args do
 			if type(arg) == "table" then
-				for key in pairs(arg) do
+				for key in arg do
```
ShadowDaughter: Same as line 291

Cmdr/Shared/Registry.luau
```diff
@@ -306,42 +309,82 @@ function Registry:RegisterCommandObject(commandObject)
...
+
+		local index = table.find(self.CommandsArray, oldCommand)
+		if index then
+			self.CommandsArray[index] = commandObject
+		else
+			table.insert(self.CommandsArray, commandObject)
+		end
+
+		if oldCommand.Aliases then
+			for _, alias in oldCommand.Aliases do
```
ShadowDaughter: Same as line 291

Cmdr/Shared/Registry.luau
```diff
@@ -306,42 +309,82 @@ function Registry:RegisterCommandObject(commandObject)
...
+	else
 		table.insert(self.CommandsArray, commandObject)
 	end
 
-	self.Commands[commandObject.Name:lower()] = commandObject
+	self.Commands[commandKey] = commandObject
 
 	if commandObject.Aliases then
-		for _, alias in pairs(commandObject.Aliases) do
+		for _, alias in commandObject.Aliases do
```
ShadowDaughter: Same as line 291

Cmdr/Shared/Registry.luau
```diff
@@ -417,9 +465,15 @@ function Registry:RegisterDefaultCommands(arrayOrFunc: { string } | (any) -> boo
 
 	local dictionary = if type(arrayOrFunc) == "table" then Util.MakeDictionary(arrayOrFunc) else nil
 
-	self:RegisterCommandsIn(self.Cmdr.DefaultCommandsFolder, dictionary ~= nil and function(command)
-		return dictionary[command.Group] or false
-	end or arrayOrFunc)
+	self:RegisterCommandsIn(
+		self.Cmdr.DefaultCommandsFolder,
+		if dictionary ~= nil
+			then function(command)
```
ShadowDaughter: This works but having a function within an if check is hard to read - Was already present in the original code but I'd rather have more readable code than saving on a few lines

ShadowDaughter (approved):
LGTM

PR #380, opened 2026-07-19
feat(client): improve client API and GUI handling

### LINKED ISSUE #298: Make "mash to enable" independent of CmdrClient:SetEnabled
Currently, the mash to enable feature will override any developer-made calls to CmdrClient:SetEnabled. For instance, to disable the console. Enabling the mash to open feature will also disable the console. Similarly, any calls to SetEnabled after SetMashToEnable is called will override mash to enable (i.e. calling SetEnabled(true) will bypass mash to enable, calling SetEnabled(false) will require mashing before the console re-opens).

This could be a problem for games which use CmdrClient:SetEnabled to restrict console access, but which also use mash to enable to avoid accidental console opening.

While this does not break any API contracts, it does change behaviour, so we should be mindful about developer and user needs here. We don't need to take any special steps to notify developers of this change in behaviour aside from a clear bullet in the changelog.

### LINKED ISSUE #310: Add client-side events to allow developers to integrate external systems into Cmdr
I propose adding the following event to the client (with a framework to add more in future):

- Cmdr.Toggled: RBXScriptSignal -> (visible: boolean)

This event would allow for developers to let their game systems, especially their UI, behave appropriately with the Cmdr client.

### LINKED ISSUE #354: Add CmdrClient property for the gui
Title. This will make it easier for developers to modify the Gui and entirely eliminate [this](https://github.com/evaera/Cmdr/issues/125#issuecomment-1622393784) problem (where in some circumstances there can be a race condition, so a dev would have to either deal with their modifications not applying, or do a `for` loop to ensure their modifications hit any possible gui)

ShadowDaughter:
Improves the `CmdrClient` API and GUI handling by moving more responsibility to the client and exposing additional functionality for developers.

The Cmdr GUI is now created on the client, allowing `CdmrClient.Gui` to provide direct access to the current GUI instance. This also fixes issues with duplicate GUI instances being created and resolves several textbox capture/focus issues.

The mash activation system has also been reworked by replacing `SetMashToEnable` with `SetMashToOpen`. This better represents its purpose as an activation method rather than a way of controlling whether Cmdr is enabled, preventing it from overriding developer calls to `CmdrClient:SetEnabled`. The old method still exists for backwards compatability but was marked as deprecated.

Additionally, this adds the `CmdrClient.Toggeld` signal for listening to visibility changes and introduces a new `exit` command. The `clean` command has been updated to make use of the new `CmdrClient.Gui` property.

Closes #298, #310, #354

## Test Code
```lua
local Players = game:GetService("Players")

local Cmdr = require(ReplicatedStorage:WaitForChild("CmdrClient"))

local Player = Players.LocalPlayer
local PlayerGui = Player:WaitForChild("PlayerGui")

local CmdrGui = Cmdr.Gui
local ButtonGui = Instance.new("ScreenGui")
local OpenButton = Instance.new("TextButton")

Cmdr:SetActivationKeys({ Enum.KeyCode.F2 })
Cmdr:SetMashToOpen(true)
Cmdr.Toggled:Connect(function(visible: boolean)
	print(visible)
end)

ButtonGui.Parent = PlayerGui
OpenButton.Parent = ButtonGui
OpenButton.Size = UDim2.fromOffset(200, 50)
OpenButton.BackgroundColor3 = Color3.new(1, 1, 1)
OpenButton.MouseButton1Click:Connect(function()
	Cmdr:Toggle()
end)

CmdrGui.Frame.Entry.TextLabel.TextColor3 = Color3.new(0, 1, 0)
```

## Declarations

### DIFF
--- a/Cmdr/BuiltInCommands/Utility/clear.luau
+++ b/Cmdr/BuiltInCommands/Utility/clear.luau
@@ -1,23 +1,19 @@
-local Players = game:GetService("Players")
-
 return {
 	Name = "clear",
 	Aliases = {},
 	Description = "Clear all lines above the entry line of the Cmdr window.",
 	Group = "DefaultUtil",
 	Args = {},
-	ClientRun = function()
-		local player = Players.LocalPlayer
-		local gui = player:WaitForChild("PlayerGui"):WaitForChild("Cmdr")
-		local frame = gui:WaitForChild("Frame")
-
-		if gui and frame then
+	ClientRun = function(context)
+		local frame = context.Cmdr.Gui:WaitForChild("Frame")
+		if frame then
 			for _, child in pairs(frame:GetChildren()) do
 				if child.Name == "Line" and child:IsA("TextBox") then
 					child:Destroy()
 				end
 			end
 		end
+
 		return ""
 	end,
 }
--- a/Cmdr/BuiltInCommands/Utility/exit.luau
+++ b/Cmdr/BuiltInCommands/Utility/exit.luau
@@ -1,0 +1,11 @@
+return {
+	Name = "exit",
+	Aliases = {},
+	Description = "Exit the Cmdr window.",
+	Group = "DefaultUtil",
+	Args = {},
+	ClientRun = function(context)
+		context.Cmdr:Hide()
+		return ""
+	end,
+}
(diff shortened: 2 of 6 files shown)


wilyt1 (approved):

PR #384, opened 2026-07-19
feat: implement Deprecation Warnings

### LINKED ISSUE #305: Deprecation warnings
Cmdr itself is over five years old, some of the older parts of the code (like the interface) are even older. By this time next month, it'll have been half a decade since Cmdr 1.0.0 released.

Features have been added, removed and changed. In order to sustain our commitment to Cmdr's long-term reliability and minimise any pain that may occur in the unlikely event of a major update (2.0.0), we need a reliable system to allow developers to be warned when they're using outdated (deprecated) methods or relying on deprecated behaviour.

Deprecation warnings will:

- allow developers to know when they're using methods or behaviour which could be broken in a major update
- allow us to better identify what can be removed in a major update
- discourage developers from using deprecated features

## How deprecation warnings work

It's TBC which component will be responsible for handling them, probably `Util`, but there'll be private methods to register and to emit a deprecation warning. Deprecation warnings must be registered without delay to prevent false errors in developer code.

Deprecation warnings should be descriptive and written in PascalCase, for example `AddHookMethod`. They may be then accompanied by a longer description, which'll be emitted in the warning.

Example:

```lua
Util:RegisterDeprecationWarning("AddHookMethod", "This method was renamed to RegisterHook in v1.1.2. The old name exists for backwards compatibility and should not be used for new work.")

function Registry:AddHook(...)
	Util:EmitDeprecationWarning("AddHookMethod")
	Registry:RegisterHook(...)
end
```

This would present in the output:

```
[Cmdr] [DeprecationWarning::AddHookMethod] This method was renamed to RegisterHook in v1.1.2. The old name exists for backwards compatibility and should not be used for new work.
(stack trace to Registry:AddHook)
```

## Suppression of deprecation warnings

A `Util:SuppressDeprecationWarning` method would exist to silence a deprecation warning. A deprecation warning may also be silenced by default if - for instance - it is no longer required. This method may be proxied into the `Cmdr` and `CmdrClient` singletons (TBC).

The method would accept either a string (to suppress a single warning) or an array of strings (to suppress multiple warnings).

Example:

```lua
Util:SuppressDeprecationWarning("AddHookMethod")
Util:SuppressDeprecationWarning({ "AddHookMethod", "GetCommandsAsStringsMethod" })
```

Calling with an invalid deprecation warning name (i.e. an unregistered one) would emit an error.

## What should have deprecation warnings

As above, the `AddHook` and `GetCommandsAsStrings` methods on the `Registry` are good places to start.

The metatable 'shadowing' behaviour (server -> registry, client -> dispatcher) can be confusing and should be distracted; this'll also encourage developers to directly interact with the components they want to use and help reduce support requests.

We should also be on the lookout for other deprecated behaviours or methods, and ensure they're warned appropriately.

## Versioning policy

While a deprecation can technically be included within a patch, the inclusion of a deprecation warning needs to be in a minor version.

Deprecation warnings cannot be removed in a patch or minor version, only a major version, although they may be silenced by default in a patch version or greater.

With a new major version, all deprecated elements should be removed, and all deprecation warnings should be removed accordingly.
 
## Future

In the future, we could make it possible for developers to programmatically listen to deprecation warnings. Although there are currently no such plans, implementation of this would therefore need to be described and approved in a new issue ticket, after deprecation warnings have been released.

ShadowDaughter:
This implements a deprecation warning system, allowing deprecated APIs to provide clear feedback to developers while maintaining backwards compatibility. Deprecated methods can now register a warning through `Util` and emit that warning when called.

Deprecation warnings are registered with a descriptive PascalCase name and a longer message explaining the replacement. When emitted, the warning includes the name, description, and a stack trace pointing back to the deprecated method that triggered it.
Closes #305 

Example output:
```
[Cmdr] [DeprecationWarning::AddHook] This method was renamed to RegisterHook in v1.1.2. The old name exists for backwards compatibility and should not be used for new work.
ReplicatedStorage.CmdrClient.Shared.Registry:590 function AddHook
Players.KSSAdministration.PlayerScripts.Client:8
```

## Deprecation API
Deprecated methods can now be marked with `Util:_registerDeprecationWarning(<name>, <description>)` and emit a warning with `Util:_emitDeprecationWarning(<name>)`, like so:
```lua
function Registry:AddHook(...)
	Util:_emitDeprecationWarning("AddHook")
	Registry:RegisterHook(...)
end

Util:_registerDeprecationWarning(
	"AddHook",
	"This method was renamed to RegisterHook in v1.1.2. The old name exists for backwards compatibility and should not be used for new work."
)
```

The deprecated API continues to work normally, but users are informed that it should no longer be used for new code.

## Warning suppression
A new `Util:SuppressDeprecationWarning` method allows developers to silence specific deprecation warnings when needed:
```lua
Util:SuppressDeprecationWarning("SetMashToEnable")
Util:SuppressDeprecationWarning({"SetMashToEnable", "GetCommandsAsStrings"})
```

## Test Code
```lua
local CmdrClient = require(ReplicatedStorage:WaitForChild("CmdrClient"))

CmdrClient.Util:SuppressDeprecationWarning({"SetMashToEnable", "GetCommandsAsStrings"})
CmdrClient:SetMashToEnable(true) -- No warning, it was supressed!
print(CmdrClient.Registry:GetCommandsAsStrings()) -- No warning, it was also supressed!

CmdrClient.Registry:AddHook("BeforeRun", function(context) -- Deprecation Warning, it's not supressed!
	print("This hook should throw a warning!")
end)
```

## Declarations

### DIFF
--- a/Cmdr/CmdrClient/init.luau
+++ b/Cmdr/CmdrClient/init.luau
@@ -211,12 +211,18 @@ function Cmdr:SetMashToOpen(enabled: boolean)
 end
 
 --[=[
-	@method SetMashToEnable
 	@deprecated v1.13.0 -- This method was renamed to SetMashToOpen in v1.13.0. The old name exists for backwards compatibility and should not be used for new work.
-	@param enabled boolean
 	@within CmdrClient
 ]=]
-Cmdr.SetMashToEnable = Cmdr.SetMashToOpen
+function Cmdr:SetMashToEnable(enabled: boolean)
+	Util:_emitDeprecationWarning("SetMashToEnable")
+	Cmdr:SetMashToOpen(enabled)
+end
+
+Util:_registerDeprecationWarning(
+	"SetMashToEnable",
+	"This method was renamed to SetMashToOpen in v1.13.0. The old name exists for backwards compatibility and should not be used for new work."
+)
 
 --[=[
 	Sets the hide on 'lost focus' feature.
--- a/Cmdr/CmdrClient/init.luau
+++ b/Cmdr/CmdrClient/init.luau
@@ -216,7 +216,7 @@ end
 ]=]
 function Cmdr:SetMashToEnable(enabled: boolean)
 	Util:_emitDeprecationWarning("SetMashToEnable")
-	Cmdr:SetMashToOpen(enabled)
+	self:SetMashToOpen(enabled)
 end
 
 Util:_registerDeprecationWarning(
--- a/Cmdr/Shared/Registry.luau
+++ b/Cmdr/Shared/Registry.luau
@@ -586,7 +586,7 @@ end
 ]=]
 function Registry:AddHook(...)
 	Util:_emitDeprecationWarning("AddHook")
-	Registry:RegisterHook(...)
+	self:RegisterHook(...)
 end
 
 Util:_registerDeprecationWarning(
--- a/Cmdr/Shared/Registry.luau
+++ b/Cmdr/Shared/Registry.luau
@@ -382,9 +382,7 @@ function Registry:RegisterCommand(
 	local commandObject = require(commandScript)
 	assert(
 		typeof(commandObject) == "table",
-		`[Cmdr] Invalid return value from command script "{commandScript.Name}" (CommandDefinition expected, got {typeof(
-			commandObject
-		)})`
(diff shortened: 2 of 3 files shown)


wilyt1 (approved):

PR #385, opened 2026-07-19
fix: bind Command for proper TextChatService support

ShadowDaughter:
Removes LegacyChatService support from the bind command in favor of TextChatService. Also adds proper cleanup for player-bound binds when players leave and consolidates MessageReceived handling into a shared connection with the use of a helper function to prevent unnecessary event connections.

## Declarations:


wilyt1 (approved):

PR #397, opened 2026-07-23
Modernize the Cmdr codebase

### LINKED ISSUE #67: Hiding commands from the default help command
For example, hiding admin commands from the everyday user if Cmdr is being used as a gameplay element as well as an admin console. Maybe a BeforeRun hook could be used for this, maybe the help command could read from `context.State.Hidden` if it exists, or maybe `context.State.Whitelist`?

A custom help command can do this, but what about the default one?

### LINKED ISSUE #91: Change casing to follow Roblox style guide
This likely won't happen until a large revamp that pushes us into a new major version (v2).

### LINKED ISSUE #303: Autocomplete: treat aliases as commands in their own right
Currently, auto-complete will ignore aliases unless there is a direct match. This is problematic because aliases can be helpful in trying to make potentially verbose or unclear command names more memorable, and so excluding them from auto-complete is harmful to the user experience.

This is a change in behaviour, so we should be mindful about developer and user needs, but there would not be any API impact.

### LINKED ISSUE #312: Stronger typo-guarding through metatable errors
As much as possible, but especially for our main singletons (Cmdr, CmdrClient, Registry, etc) we should use metatables to guard against typos. In other words, we should ensure that we're raising errors whenever developers try to access something that doesn't exist from Cmdr. This will help developers to spot errors immediately and clearly, reducing support requests. I have a mental image of how this should work so I'll try add this when I get a minute.

ShadowDaughter:
This PR modernizes several core parts of Cmdr with a focus on readability, maintainability, and Luau type safety.
It aims to modernize the codebase by removing outdated patterns and adopting modern Luau language features. This PR is intended to be a refactor only; no user-facing behavior should change, with the exception of newly added deprecation warnings.

## Changes

- Added comprehensive Luau type annotations throughout the codebase.
- Added documentation to previously undocumented modules, particularly UI-related components.
- Refactored large portions of the codebase for improved readability and consistency.
- Replaced legacy language patterns with modern Luau equivalents (e.g. interpolated strings).
- Standardized coding style and conventions across modules.
- Performed general cleanup of legacy code and internal implementations.

## Reasoning

Cmdr has been around for many years and, while it remains a robust library, parts of the codebase have fallen behind modern Luau conventions. This PR aims to reduce technical debt by modernizing internal implementations, improving static type safety, and making the codebase easier to understand and maintain.

By adopting modern language features and improving consistency across modules, future contributions become easier to develop, review, and reason about. These changes also allow Luau's type checker to catch more issues at development time without intentionally affecting Cmdr's runtime behavior.

This should provide a solid foundation for future development releases and serve as a reliable base for future reworks across the codebase and bring us a step closer towards a potential v2.0 remake.

## Declarations


autonordev:
You might've mentioned this on chat, but is this something now ready for a review? These sorts of things were stuff I wanted in the next release. Ik I've been stepping away from hands-on coding but this definitely should get a comprehensive review from both maintainers!

ShadowDaughter:
@autonordev I've marked the PR ready for review now that we've given it some testing and ensured it worked as expected, feel free to give it a readthrough!

Cmdr/BuiltInCommands/Admin/banServer.luau
autonordev, line 18: random thought, might be worth logging as a separate ticket: I think there's a lot of use cases which are "string that is not empty" (`announce` for example). creating that as a sub-type of string, say `stringNotEmpty` would probably reduce a lot of boilerplate.

Cmdr/BuiltInCommands/Admin/kickServer.luau
```diff
@@ -18,7 +17,7 @@ return function(context, players, reason)
 		end
 	end
 
-	for _, player in pairs(players) do
+	for _, player: Player in pairs(players) do
```
autonordev:
```suggestion
	for _, player in pairs(players) do
```

Isn't type `Player` inferred by `for x in { Players }`

autonordev: Is `pairs` current good practice? I think Luau defaults to pairs if you just do `in {...}`, but this is years-old knowledge on my part. I know historically you'd do `ipairs` over arrays to be slightly faster, I remember some discussion about how the interpreter could discern whether whether a table was an array-like (so `ipairs`) or not (so `pairs`, e.g. dictionaries, mixed-keys, or unordered keys). That said, this would be micro-optimisation at best, so this comment is ok to ignore.

Cmdr/BuiltInCommands/Admin/killServer.luau
```diff
@@ -1,9 +1,13 @@
...
+		const humanoid = character and character:FindFirstChildWhichIsA("Humanoid")
+		if not humanoid then
+			continue
 		end
+
+		humanoid.Health = 0
 	end
 
-	return ("Killed %d players."):format(#players)
+	return `Killed {#players} players.`
```
autonordev, line 12: Random thought, so either ignore it or log it as another ticket, but in my own Cmdr projects I had a pluralize utility (and often just wrote this by hand) which would swap `players` with `player` if it was 1.

Cmdr/BuiltInCommands/Admin/teleportServer.luau
```diff
@@ -1,24 +1,33 @@
...
 		end
+
+		task.spawn(function()
+			if humanoid.SeatPart then
+				humanoid.Sit = false
+				humanoid:GetPropertyChangedSignal("SeatPart"):Wait()
+			end
+
+			if character:IsDescendantOf(workspace) then
+				character:PivotTo(cFrame)
```
autonordev, line 27: `PivotTo` will maintain the current behaviour but for teleporting mightn't `MoveTo` be preferable? This is what we do with `blink`. Just a thought.

> If there are any obstructions where the model is to be moved, such as Terrain or other BaseParts, the model will be moved vertically upward until there is nothing in the way. If this behavior is not desired, PVInstance:PivotTo() should be used instead.

that said I thought `MoveTo` had been deprecated in favour of `PivotTo` lol

Adding this so you can easily decide what you want:

```suggestion
				character:MoveTo(cFrame)
```

Cmdr/BuiltInCommands/Debug/version.luau
```diff
@@ -1,12 +1,13 @@
-local version = "v1.12.0"
+const VERSION = "v1.13.0-rc.2"
 
 return {
 	Name = "version",
-	Args = {},
-	Description = "Shows the current version of Cmdr",
+	Aliases = { "v" },
```
autonordev: Is this alias necessary? It can easily conflict with developer commands and user-specified aliases.

(Previously, a user-specified alias would override anyway, because newer commands override older commands and aliases are just commands, and I'd _hope_ user aliases are treated as non-builtin commands for prioritisation purposes lol. But the next change on my screen prevents aliases from overriding commands lol.)

I don't know how to tell GitHub to suggest deleting a line. I guess this might work:

```suggestion
```

Cmdr/BuiltInCommands/Utility/clear.luau
```diff
@@ -1,16 +1,18 @@
 return {
 	Name = "clear",
-	Aliases = {},
+	Aliases = { "clr" },
```
autonordev: See my comment on `version`. 

(Fun fact, here's my Cmdr user init script from the last game I worked on.)

```
alias cd tp . @226.82188415527344,2.299931049346924,-241.73904418945312
alias office tp . @1214.02783203125,220.08404541015625,-1863.5556640625
alias vpd tp . @-166.3881378173828,1.6499996185302734,-1123.794189453125
alias bcss tp . @1045.29638671875,1.6499996185302734,-2467.453369140625
alias ps tp . @395.2143859863281,1.6499998569488525,-662.1932373046875
alias ylcs tp . @538.852783203125,1.6499137878417969,-722.5415649414062
alias cfb tp . @-1301.8345947265625,1.2808470726013184,992.7081298828125
alias rcmp tp . @-215.1818389892578,44.69718933105469,7998.99560546875
alias sid tp . @1054.09521484375,1.6500005722045898,-313.64691162109375
alias doba tp . @1494.5880126953125,3.6146130561828613,-1003.6593627929688
alias bcas tp . @1183.2547607421875,1.6499998569488525,-1195.9290771484375
alias vfrs tp . @450.7861022949219,1.6454708576202393,-2256.5380859375
alias bcaa tp . @1373.95947265625,1.6499998569488525,-1714.757080078125
bind N blink
bind J thru
alias h setname . hide
alias s setname .
alias vc v && c
alias c clear
alias uc "removetool . "LE | C8"" && "removetool . "LE | Remington 870""
```

*(Note that penultimate alias :P)*

```suggestion
```

Cmdr/BuiltInCommands/help.luau
```diff
@@ -8,7 +11,7 @@ Argument Shorthands
 ?N  List of N random values
 ]]
 
-local TIPS = [[
+const TIPS = [[
 Tips
 ----
 • Utilize the Tab key to automatically complete commands
```
autonordev:
```suggestion
• Use the Tab key to autocomplete commands and suggestions
```

I originally went with `values` instead of `suggestions` but I think suggestions works better. Especially now we're in the AI era where tab suggestions are in all sorts of apps. (Autocomplete is also a well-understood word, see: every modern smartphone or code editor ever.)

Cmdr/BuiltInCommands/Admin/gotoPlace.luau
```diff
@@ -1,27 +1,26 @@
 return {
 	Name = "goto-place",
-	Aliases = {},
-	Description = "Teleport to a Roblox place",
+	Description = "Teleport to a Roblox place.",
 	Group = "DefaultAdmin",
 	AutoExec = {
-		'alias "follow-player|Join a player in another server" goto-place $1{players|Players} ${{get-player-place-instance $2{playerId|Target}}}',
-		'alias "rejoin|Rejoin this place. You might end up in a different server." goto-place $1{players|Players} ${get-player-place-instance ${me} PlaceId}',
+		'alias "follow-player|Join a player in another server." goto-place $1{players|Players|The players you want to teleport.} ${{get-player-place-instance $2{playerId|Target|The player you want to follow.}}}',
```
autonordev:
```suggestion
		'alias "follow-player|Join a player in another server." goto-place $1{players|Players|The players to teleport.} ${{get-player-place-instance $2{playerId|Target|The player you want to follow.}}}',
```

Conciseness (and consistency)

Cmdr/BuiltInCommands/Admin/gotoPlace.luau
```diff
@@ -1,27 +1,26 @@
...
+		'alias "rejoin|Rejoin this place. You might end up in a different server." goto-place $1{players|Players|The players to teleport.} ${get-player-place-instance ${me} PlaceId}',
 	},
 	Args = {
 		{
 			Type = "players",
 			Name = "Players",
-			Description = "The players you want to teleport",
+			Description = "The players you want to teleport.",
```
autonordev:
```suggestion
			Description = "The players to teleport.",
```

same as R6

Cmdr/BuiltInCommands/Admin/gotoPlace.luau
```diff
@@ -1,27 +1,26 @@
...
 		},
 		{
 			Type = "positiveInteger",
 			Name = "Place ID",
-			Description = "The Place ID you want to teleport to",
+			Description = "The Place ID you want to teleport to.",
```
autonordev:
```suggestion
			Description = "The Place ID to teleport to.",
```

same as R6

Cmdr/BuiltInCommands/Admin/gotoPlace.luau
```diff
@@ -1,27 +1,26 @@
...
 		},
 		{
 			Type = "string",
 			Name = "JobId",
-			Description = "The specific JobId you want to teleport to",
+			Description = "The specific JobId you want to teleport to.",
```
autonordev:
```suggestion
			Description = "The JobId to teleport to.",
```

Cmdr/BuiltInCommands/Admin/kick.luau
```diff
@@ -7,12 +7,12 @@ return {
...
 			Type = "players",
 			Name = "Players",
-			Description = "The players to kick",
+			Description = "The players to kick.",
 		},
 		{
 			Type = "string",
 			Name = "Reason",
-			Description = "Kick message shown to the player",
+			Description = "Kick message shown to the player.",
```
autonordev:
```suggestion
			Description = "Kick message shown to the players.",
```

Cmdr/BuiltInCommands/Admin/unban.luau
```diff
@@ -6,12 +6,12 @@ return {
 		{
 			Type = "playerIds",
 			Name = "Players",
-			Description = "The players to unban (max 50)",
+			Description = "The players to unban (max 50).",
 		},
 		{
 			Type = "boolean",
 			Name = "Apply to Universe",
```
autonordev:
```suggestion
			Name = "Apply to universe",
```

NB: Roblox doesn't call them Universes anymore

Cmdr/BuiltInCommands/Admin/ban.luau
```diff
@@ -6,40 +6,40 @@ return {
...
 		},
 		{
 			Type = "duration",
 			Name = "Duration",
-			Description = "How long the ban lasts (negative values indicate a permanent ban)",
+			Description = "How long the ban lasts (negative values indicate a permanent ban).",
 		},
 		{
 			Type = "string",
 			Name = "Display Reason",
```
autonordev:
```suggestion
			Name = "Display reason",
```

Title Case is evil. If I ever wrote a Cmdr command/text style-guide, this would be rule 0.

Roblox calls this the _public reason_, we could go with that for consistency:

```suggestion
			Name = "Public reason",
```

I should've flagged these on the ban PR itself. But this is a megapr so it's an ok vehicle for it.

Cmdr/BuiltInCommands/Admin/ban.luau
```diff
@@ -6,40 +6,40 @@ return {
...
-			Description = "Ban message shown to the player",
+			Description = "Ban message shown to the player.",
 		},
 		{
 			Type = "string",
 			Name = "Private Reason",
```
autonordev:
```suggestion
			Name = "Private reason",
```

Cmdr/BuiltInCommands/Admin/ban.luau
```diff
@@ -6,40 +6,40 @@ return {
...
-			Description = "Internal staff notes for ban history",
+			Description = "Internal staff notes for ban history.",
 			Optional = true,
 		},
 		{
 			Type = "boolean",
 			Name = "Exclude Alt Accounts",
```
autonordev:
```suggestion
			Name = "Exclude alt accounts",
```

Cmdr/BuiltInCommands/Admin/ban.luau
```diff
@@ -6,40 +6,40 @@ return {
...
-			Description = "Skip banning their alternate accounts",
+			Description = "Skip banning their alternate accounts.",
 			Default = false,
 		},
 		{
 			Type = "boolean",
 			Name = "Apply to Universe",
```
autonordev:
```suggestion
			Name = "Apply to universe",
```

Cmdr/BuiltInCommands/Admin/ban.luau
```diff
@@ -6,40 +6,40 @@ return {
...
-			Description = "Ban from all places in this experience",
+			Description = "Ban from all places in this experience.",
 			Default = true,
 		},
 		{
 			Type = "boolean",
 			Name = "Apply Device Block",
```
autonordev:
```suggestion
			Name = "Apply device block",
```

Cmdr/BuiltInCommands/Utility/alias.luau
```diff
@@ -16,9 +15,12 @@ return {
 		},
 	},
 
-	ClientRun = function(context, name, commandString)
-		context.Cmdr.Registry:RegisterCommandObject(context.Cmdr.Util.MakeAliasCommand(name, commandString), true)
+	ClientRun = function(context: any, name: string, commandString: string): string
+		if context.Cmdr.Registry._commands[name] then
+			return `There is already a command registered with the name {name}.`
+		end
```
autonordev: This is a behaviour change isn't it? I'm on the fence but I can see aliases overriding commands causing debug headaches. But, especially some builtin aliases, might want to be overridden by end users (if not overridden by developers).

Dealer's choice:

```suggestion
```

Cmdr/BuiltInCommands/Utility/history.luau
```diff
@@ -12,12 +11,12 @@ return {
 		{
 			Type = "integer",
 			Name = "Line Number",
-			Description = "Command line number (can be negative to go from end)",
+			Description = "Command line number (can be negative to go from end.)",
```
autonordev:
```suggestion
			Description = "Command line number (can be negative to go from end)",
```

rmv the ending period (American English is wrong on this anyway, but visually a period around brackets, especially if the sentence has no other punctuation, is noisy.)

Cmdr/BuiltInTypes/Vector.luau
```diff
@@ -1,59 +1,55 @@
-local Util = require(script.Parent.Parent.Shared.Util)
+const AXES: { string } = { "X", "Y", "Z" }
```
autonordev, line 1: is `{string}` not implied?

Cmdr/CmdrClient/Shared/Command.luau
autonordev, line 1: I get that the point here is to avoid having to shuffle modules about but I feel like there's a better way instead of the server reaching into CmdrClient for `Shared`.

Cmdr/init.luau
```diff
@@ -1,10 +1,21 @@
...
+const ReplicatedStorage = game:GetService("ReplicatedStorage")
+const RunService = game:GetService("RunService")
 
-if RunService:IsServer() == false then
-	error(
-		"[Cmdr] Client scripts cannot require the server library. Please require the client library from the client to use Cmdr in your own code."
-	)
+const IS_SERVER_CONTAINER = script:FindFirstAncestorWhichIsA("ServerScriptService") or script:FindFirstAncestorWhichIsA("ServerStorage")
+const IS_SERVER = RunService:IsServer()
+assert(IS_SERVER_CONTAINER, "[Cmdr] The Cmdr server library must be kept within a Server Container.")
```
autonordev:
```suggestion
if not IS_SERVER_CONTAINER then
	warn("[Cmdr] The package should be kept within a server-side location. Future versions of Cmdr will require the server library be kept within ServerScriptService or ServerStorage.")
end
```

!block! I'm not sure this should be a full error, as this might break some niche setups. This is behaviour we should deprecate first. While Cmdr currently behaves unexpectedly outside of a server container, this is a breaking change.

docs/03-commands-reference/01-commands.md
```diff
@@ -36,19 +36,19 @@ The implementation — whether on the server or client — is passed the [Comman
 ```lua title="TeleportServer.luau"
 -- These arguments are guaranteed to exist and be correctly typed.
 return function (context, fromPlayers, toPlayer)
-  if toPlayer.Character and toPlayer:FindFirstChild("HumanoidRootPart") then
-    local position = toPlayer.Character.HumanoidRootPart.CFrame
+	if toPlayer.Character and toPlayer:FindFirstChild("HumanoidRootPart") then
```
autonordev, line 39: IIRC there was a reason I used spaces on the website. I think Docusaurus and browsers behave weirdly with `\t` characters.

docs/95-community/01-cookbook.md
```diff
@@ -74,7 +74,7 @@ local function filterText(text, fromUserId)
 	return broadcastSuccess and filteredString
 end
 
-return function(context, targets, title, message, duration)
+return function(context: any, targets, title, message, duration)
```
autonordev: I can't instantly glean what the types here should be, but this commit should add them, _or_ an issue should be created for them to be added. This shouldn't just get lost.

CHANGELOG.md
```diff
@@ -103,7 +103,7 @@
 - [Aliases can now include type checking and name information in arguments](https://eryn.io/Cmdr/guide/MetaCommands.html#alias) (#56)
 - New default command: "me". Displays the current player's name.
 - Make `bring` and `to` commands built-in aliases for `teleport`. `to` has effectively moved from the `DefaultDebug` group to the `DefaultAdmin` group. (#115)
-- The default activation key is now <kbd>F2</kbd>
+- The default activation key is now `F2`
```
autonordev: What's wrong with `<kbd>` <kbd>F2</kbd> `</kbd>`

.markdownlint.yaml
autonordev, line 1: Could markdownlint be added to `rokit` or `package.json`?

(I know adding it to Rokit sounds rogue, but I managed to get `dotenvx` as a Rokit dependency. _Unfortunately, `just` didn't work as one._)

stylua.toml
```diff
@@ -1,4 +1,4 @@
-column_width = 120
+column_width = 200
```
autonordev: !block! What's the motivation behind this change? 200 columns is a _lot_. Only ~140 fit on my 13" MacBook (and that's with just one standard sidebar open).

Every style I've ever seen or written has commanded keeping code visible in most instances, without needing to scroll.

autonordev (changes requested):
Firstly: thank you so much @ShadowDaughter for what is a really comprehensive and overwhelmingly positive contribution. A lot of this is stuff I've been thinking about for years, but just haven't had the time to sit down and do.

I've read as much of the patch as I can, but some of it I just skimmed; I skipped most of the core library changes (especially CmdrInterface).

Attached are a whole host of comments. Some are thoughts, some are questions, some are suggestions (and some of them are very loose while others are very adamant). Don't just blindly click 'accept all suggestions', because some are mutually-exclusive options (accept one or neither) whereas others are optional changes (think about whether they should be accepted). Some of my suggested changes should probably _not_ be accepted, at least not for this PR.

I'd like to click <kbd>Approve</kbd> and ::shrug:: on the changes I have opinions on, but there's two things that I think need a change or further discussion (which I've prefixed with `!block!`). I'm happy to re-review and approve an updated patch, and ofc I'm happy to discuss some of these comments.

Broadly, my comments cover:

- conciseness and consistency within strings
- questions about specific Luau typing cases and current good practice
- a few instances of behaviour change, where we might want to take a different path than your patch

For simple things, especially text style nits, I've used suggested changes as much as possible, so you should be able to just accept those and move on (you can "batch" them into one single commit, too; I think it's an interactive session like the review xp). For questions, obviously I wish it was that simple! (And for random thoughts, you can just ignore them if you want, up to you.)

_P.S. GitHub should email me your replies, but I might miss them. You can just pester me on Discord to take a look. (Of course, we can also chat about the pr on discord, but GitHub is easier for me to actually digest, think, and write.)_

cc: @wilyt1 

Thank you both for your continued hard work in breathing new life into this brilliant project 💜

Cmdr/BuiltInCommands/Admin/banServer.luau
ShadowDaughter: This would be easily implementable, but I believe it's a bit niche to. It's just like shown in this code a maximum of 3 lines of code a developer would have to include. If planned to be implemented, I do think a seperate issue should be opened for this and imcluded in a future version instead.

Cmdr/BuiltInCommands/Admin/kickServer.luau
ShadowDaughter: I forgot to send my reply, but I agree that this is redundant. I will make a seperate commit to remove all instances of redundant annotation like this.

ShadowDaughter: Luau does default to pairs by simpy using `in {...}`, but I personally believe that doing this makes the code harder to read since you tend to second guess yourself when reading it to confirm if it's actually using pair or not. However, it is correct that ipairs is actually faster in cases like these which is why they are used in other instances. (Or atleast used to be? I'm not sure if they were accidentally changed to pairs at some point.)

Cmdr/BuiltInCommands/Admin/killServer.luau
ShadowDaughter: I do believe this would be better suited for a own issue and included in a future-release. For my own projects I tend to use table.concat to display a list of all target names instead, but I'm not sure if this is optimal for most uses since it might result in incredibly long outputs depending on how many targets the command was used on.

Cmdr/BuiltInCommands/Admin/teleportServer.luau
ShadowDaughter: I believe `PivotTo` is better for this usage. When I teleport somewhere, I tend to actually want to be teleported where I specified, not above or nearby just because a wall is currently in the way. This is especially true when I am teleporting to another player. While I agree that it *could* be useful to use MoveTo for general Vector3 teleportation, I don't think it's the wanted behavior here.

I will however wait for @wilyt1 's and yours opinion on this topic.

Cmdr/BuiltInCommands/Debug/version.luau
ShadowDaughter: I agree that this conflicting with developer or user-specified aliases is an issue that I haven't thought of. It however would rather be important to ensure these aliases are overriden by developer or user-specified aliases rather than not including them at all.

Will wait on @wilyt1 's and yours thoughts on this.

Cmdr/BuiltInCommands/Utility/clear.luau
ShadowDaughter: Same comment as for version.luau

Cmdr/BuiltInCommands/Admin/gotoPlace.luau
ShadowDaughter: Thank you for this, I tried going over all commands and doing so everywhere possible but it seems I missed a few spots.

Cmdr/BuiltInCommands/Admin/unban.luau
ShadowDaughter: The boolean is still called ApplyToUniverse within BanAsync, so I'd keep it for consistency with what the documentation says. Agree with the casing tho.

Cmdr/BuiltInCommands/Utility/alias.luau
ShadowDaughter: I can't fully remember why this was the agreed upon design, but I can't find the conversation either due to the mess up from Pod... I don't think an alias should be able to overwrite an existing command. Aliases should be overriden by other aliases, not an entire command.

I'll wait for your thoughts on this one again.

Cmdr/BuiltInTypes/Vector.luau
ShadowDaughter: I'm not sure if relying on it being implied is the best choice here. We want to ensure all items here are strings. If we rely in it being implied and one would were to add a number it would become `string | number`. While nobody *should* be adding a number (or modifying this variable at all), I don't believe this to be good.

Cmdr/CmdrClient/Shared/Command.luau
ShadowDaughter: The server *already* reaches into CmdrClient to access shared. Originally, shared is moved into CmdrClient during runtime and the servers access to the shared folder happens before it is moved. However, CmdrClient won't have type annotations if we rely on WaitForChild to access Shared. Keeping shared within CmdrClient from the beginning is a better and more optimal change. I don't really think we should be moving it during runtime at all since it seems like useless and redundant behavior to me when we can just keep it there from the start.

Cmdr/init.luau
ShadowDaughter: You are right that erroring here might break niche setups. We should go with the current behavior of throwing a deprecation warning before simply changing behavior entirely.

docs/03-commands-reference/01-commands.md
ShadowDaughter: I originally made this change because my markdown linter screams at using spaces over tabs. However in either case, the spaces should be 4 indents long. I will keep current behavior for consistency across the documentation since all other places already use tabs.

docs/95-community/01-cookbook.md
ShadowDaughter: Leaving this up to @wilyt1 to implement (I dislike touching the documentation)

CHANGELOG.md
ShadowDaughter: Change from my markdown linter, can be ignored or reverted if necessary. Maintainers choice.

.markdownlint.yaml
ShadowDaughter: This is just for my personal use and absolutely not even necessary to add at all. I've however made multiple changes to markdown files already relying on my markdown linter. As before, maintainers choice.

stylua.toml
ShadowDaughter: I tend to use 200 on my projects since myself and all developers I tend to work with have wider monitors, but I can see why this is *way* too much for smaller ones such as laptops or macs. Will change it back to 120.

ShadowDaughter: Adding onto this, I do believe it would be worth to go back over the code and rewrite certain lines that are split into multiple to rely on additional variables for readability. A single line of code that is so long it gets split up into multiple by the formatter is generally bad design.

Cmdr/BuiltInCommands/Admin/kickServer.luau
autonordev: NB: https://luau.org/performance/#optimized-table-iteration

Cmdr/BuiltInCommands/Admin/teleportServer.luau
autonordev: I think most admin tools use absolute position (i.e. PivotTo) instead of MoveTo. Thinking back to how many times I've tried to teleport myself and got stuck in a wall. If I was making something from scratch I'd probably prefer MoveTo but PivotTo is the existing behaviour.

Cmdr/BuiltInCommands/Debug/version.luau
autonordev: Yes I think builtin should always be overridden but I don't think adding short or ambiguous aliases like this are something we should be doing. Especially not for a command as little-used as version.

Cmdr/BuiltInCommands/Utility/alias.luau
autonordev: I suppose we could do a warning? 🤷‍♀️ I don't have a clear opinion on this, and it's a bit difficult making judgements like this when I'm no-longer using Cmdr regularly.

@wilyt1 What do you think?

autonordev (approved):
LGTM :shipit:

autonordev:
Again thank you so much for this hugely comprehensive and amazing contribution Hakai

PR #410, opened 2026-08-13
Add type definitions to cookbook

wilyt1:
Adds new definitions section to cookbook with example.

**Declarations**:


autonordev (commented):
Random thought, we could probably in the future do some voodoo with Luau-lsp's global types feature, _and_ auto-generate typedefs via Moonwave-extractor's data. (But it'd definitely be a bit weird and require a lot of work -- like how would it know how to type CommandDef or server command handlers, and obv the typedef file would need to be included in every environment to prevent errors).

Ok that random thought was actually a bad idea.

PR #414, opened 2026-08-16
Re-include executor in broadcast args

NinjaMandalorian:
With the refactor to context:BroadcastEvent, context.Executor was lost. This makes the current docs outdated, which suggest you can override the "Message" and get the player who broadcasted.

Between changing the docs or the broadcast, it would make more sense this way, as including the player by-default incurs minimal cost and ensures backwards compatibility.

**Declarations**:


autonordev:
[This commit](https://github.com/evaera/Cmdr/commit/6438f447385e6e262cecf5700a401e2c7b10397b) accidentally removed it, nice catch.

Thanks for your contribution.

PR #421, opened 2026-08-16
Re-add calls and hashtag to `var` commands

### LINKED ISSUE #419: `var` command crash bug
The guards update has used them for datastore checking. Slight issue: this will crash if datastores are disabled.

Image

wilyt1:
Closes #419. While GetDataStore emits a Studio error for published games with API access off, it emits a fatal error when used in a local place file. Also, the hashtag is required in error messages as `run-lines` skips any lines starting with them.

**Declarations**:

### DIFF
--- a/Cmdr/BuiltInCommands/Utility/varServer.luau
+++ b/Cmdr/BuiltInCommands/Utility/varServer.luau
@@ -1,26 +1,4 @@
-const DataStoreService = game:GetService("DataStoreService")
-
-const queue = {}
-local DataStoresActive, DataStore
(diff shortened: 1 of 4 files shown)


ShadowDaughter:
The functionality seems good to me, but the code being constantly reused like this feels a little off. While not necessarily required, maybe some ability to reuse this code rather than having it copy-pasted throughout 3 files is good. One approach could be having an additional function within the guard that we could call within the implementations directly, but this is entirely optional.

ShadowDaughter (approved):
LGTM!

PR #429, opened 2026-08-24
Improve `ProxyCalls` deprecation warning

### LINKED ISSUE #415: `APIUsage::ProxyCalls` error should be made clearer

autonordev:
Currently `APIUsage::ProxyCalls` exists as a special case warning, instead of being a deprecation warning. This replaces that special case with a standard deprecation warning, and makes the wording clearer.

Well, an attempt has been made to make the warning clearer. I'm not sure what exactly would be the *perfect* wording here, but this at least makes it clear that it's a deprecated thing. I think further improvement is necessary (and this should be made to the patch, and obviously in that case edit this bit of the commit message accordingly).

Closes #415

@ShadowDaughter Your rv would be appreciated here. 

**Declarations**:

### DIFF
--- a/Cmdr/CmdrClient/Shared/Util.luau
+++ b/Cmdr/CmdrClient/Shared/Util.luau
@@ -869,4 +869,11 @@ function Util:SuppressDeprecationWarning(names: string | { string })
 	end
 end
 
+Util:_registerDeprecationWarning(
+	"ProxyCalls",
+	`Method calls should be performed against their respective modules rather than the {if RunService:IsServer()
+		then "Cmdr"
+		else "CmdrClient"} object. Proxy calls are deprecated, they exist for backwards compatibility and should not be used for new work.`
+)
+
 return Util
--- a/Cmdr/CmdrClient/init.luau
+++ b/Cmdr/CmdrClient/init.luau
@@ -362,15 +362,7 @@ return setmetatable(CmdrClient, {
 		assert(value, `[Cmdr] Dispatcher has no method or property "{key}".`)
 
 		if type(value) == "function" then
-			if not script:GetAttribute("HasShownProxyCallWarning") then
-				warn(
-					debug.traceback(
-						`[Cmdr] [APIUsage::ProxyCalls] Method calls should be performed by their respective modules rather than the Cmdr object.`,
-						2
-					)
-				)
-				script:SetAttribute("HasShownProxyCallWarning", true)
-			end
+			Util:_emitDeprecationWarning("ProxyCalls")
 
 			return function(_, ...)
 				return value(dispatcher, ...)
--- a/Cmdr/init.luau
+++ b/Cmdr/init.luau
@@ -88,7 +88,7 @@ end
 	@private
 
 	@prop _replicatedRoot ModuleScript
-	
+
 	Refers to the CmdrClient modulescript containing all shared functionality.
 ]=]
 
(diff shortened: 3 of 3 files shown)


Cmdr/CmdrClient/init.luau
```diff
@@ -16,6 +16,11 @@ const DefaultEventHandlers = require("@self/DefaultEventHandlers")
 const TYPES_FOLDER = script:WaitForChild("Types") :: Folder
 const COMMANDS_FOLDER = script:WaitForChild("Commands") :: Folder
 
+Util:_registerDeprecationWarning(
+	"ProxyCalls",
+	"Method calls should be performed by their respective modules rather than the CmdrClient object. Proxy calls are deprecated, they exist for backwards compatibility and should not be used for new work."
+)
+
```
ShadowDaughter: Replace `object` with `singleton` for consistency.
```suggestion
Util:_registerDeprecationWarning(
	"ProxyCalls",
	"Method calls should be performed by their respective modules rather than the CmdrClient singleton. Proxy calls are deprecated, they exist for backwards compatibility and should not be used for new work."
)

```

ShadowDaughter: Should we potentially register this deprecation warning directly within Util? I'm not sure if registering it individually for both the server and the client is the best choice here, especially since we will have to update both warnings if we ever change it

Cmdr/init.luau
```diff
@@ -20,6 +20,11 @@ const Dispatcher = require("@self/CmdrClient/Shared/Dispatcher")
 const Registry = require("@self/CmdrClient/Shared/Registry")
 const Util = require("@self/CmdrClient/Shared/Util")
 
+Util:_registerDeprecationWarning(
+	"ProxyCalls",
+	"Method calls should be performed by their respective modules rather than the Cmdr object. Proxy calls are deprecated, they exist for backwards compatibility and should not be used for new work."
+)
+
```
ShadowDaughter: Same changes as to CmdrClient.
```suggestion
Util:_registerDeprecationWarning(
	"ProxyCalls",
	"Method calls should be performed by their respective modules rather than the Cmdr singleton. Proxy calls are deprecated, they exist for backwards compatibility and should not be used for new work."
)

```

ShadowDaughter (dismissed):
LGTM overall, and I'm not really sure how we could make this warning any clearer. I think the new wording is ok as is and doesn't necessarily need to be improved any.

Cmdr/CmdrClient/init.luau
autonordev: I prefer plain language, especially for international users (singleton doesn't easily translate, while object does), especially for a warning that is likely to hit a lot of people... but I see the consistency point. @wilyt1 What do you think?

autonordev: Good idea. I'll update the patch when I get a chance.

Cmdr/CmdrClient/Shared/Util.luau
```diff
@@ -869,4 +869,11 @@ function Util:SuppressDeprecationWarning(names: string | { string })
 	end
 end
 
+Util:_registerDeprecationWarning(
+	"ProxyCalls",
+	`Method calls should be performed against their respective modules rather than the {if RunService:IsServer()
+		then "Cmdr"
+		else "CmdrClient"} object. Proxy calls are deprecated, they exist for backwards compatibility and should not be used for new work.`
+)
+
```
ShadowDaughter, line 878: Not really necessary but maybe create a `IS_SERVER` variable? We already check if we are on the server at the top of the module for datastore purposes

Cmdr/CmdrClient/init.luau
wilyt1: I agree with keeping object and updating other warnings. Singleton is just a specialized architectural pattern, and we should keep the language as plain and approachable as possible for the broader audience.

PR #431, opened 2026-08-24
Update CHANGELOG.md

### LINKED ISSUE #418: Release notes need to be updated
There are a few minor things that were in the RC announcements but not mentioned in the changelog.

This includes things like ProxyCall deprecation warning, and the replicated container deprecation warning (the latter I'm pretty sure is in there already)

(I'll handle this one before cutting the next release.)

autonordev:
* Tidy some things up
* Add references to a few unreported changes, including the new keybinds
* Explicitly call out new errors and ProxyCalls deprecation

(This is also a good opportunity to suggest any other changes to the 1.13 changelog. Obviously, update the commit message to not include this note and, if substantive changes have been made, to mention those above!)

Closes #418

**Declarations**:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -47,7 +47,7 @@ This type allows you to use `~[studs]` (where `[studs]` is a number of studs) fo
 
 **Deprecation warnings**
 
-Cmdr is over 8 years old, and we maintain as much compatibility as possible with previous versions. In order to avoid future pain, we've added deprecation warnings, for example when using the old name of a method.
+Cmdr is over 8 years old, and we maintain as much compatibility as possible with previous versions. In order to avoid future pain, we've added deprecation warnings, for example when using the old name of a method or when using proxy calls like `Cmdr:RegisterCommandsIn` instead of `Cmdr.Registry:RegisterCommandsIn`.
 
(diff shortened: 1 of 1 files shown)


CHANGELOG.md
```diff
@@ -100,6 +102,9 @@ Deprecation warnings are emitted with the following format `[Cmdr] [DeprecationW
 
 - [Fixed replication bugs causing commands or types to not load on the client.](https://github.com/evaera/Cmdr/commit/8db4824f52bef023474c837fb2667a0b359a69df)
 
+- Command aliases now display correctly with autocomplete.
+
+- 3 keybinds have been added: `ESC` to close the UI, `LCTRL` + `BACKSP` to clear the input line, and `TAB` to recapture input line focus.
```
ShadowDaughter, line 107: Something else that might be worth noting: Cmdr used to automatically recapture textbox focus whenever you clicked anywhere within the Cmdr UI. I removed this in favor of using `TAB` to capture focus, as the previous behavior prevented you from copying any output from the UI. This was a personal issue I encountered frequently, which also made some commands that returned output effectively obsolete, since their output couldn't be copied.

While I believe for this to be a good change, it does change existing behavior and should most likely be noted here.

ShadowDaughter (commented):
One small note, rest LGTM!

CHANGELOG.md
autonordev: I never encountered this issue you describe. And I don't think it'd be expected behaviour anyway, let alone desired behaviour. So I think we're OK.

wilyt1 (approved):
Looks good to me!

PR #432, opened 2026-08-24
Restructure documentation articles

### LINKED ISSUE #430: ix structure
Current structure:

- Introduction
- Getting Started (my initial thinking is to expand this out)
	- Installation
	- Set-up
	- Updating
- Commands & Reference (my initial thinking is to expand this out too)
	- Commands
	- Meta-commands (move to Advanced)
	- Hooks
	- Types
	- Network event handlers (move to Advanced)
	- Automatic execution (move to Advanced)
- Advanced (rename to Guides, move out Permissions and place it one below Hooks)
 	- delete Best practice
- Community (should probably be split out, with Cookbook and securityreport being standalone or miscellanea; cookbook can be standalone or moved to advanced)

autonordev:
We've tried a few different layouts for our documentation, and I experimented a bit with this patch, but ultimately there doesn't seem to be anything cleaner than just having it all flat (which is also what the legacy site did, we just have more content than it ever did).

This change also introduces redirects, including for the now-clobbered paths as well as previous changes and tweaks that we made to the site.

Closes #430

**Declarations**:


ShadowDaughter:
I quickly pulled your commits and gave it a look and I'm not sure if I really like this change.
While it in theory makes everything more accessible, it also causes the documentation to feel incredibly unorganized and unstructured. The provided indexing also makes no real sense, why is updating just in the middle of the list rather than at the top? And as stated before, I'm not too sure if a whole guide on updating is even necessary when it is essentially very similar to a regular installation of Cmdr.

wilyt1:
What if we merge the contents of the Updating page into installation?

autonordev:
> What if we merge the contents of the Updating page into installation?

> I'm not too sure if a whole guide on updating is even necessary when it is essentially very similar to a regular installation of Cmdr.

I think updating being kept separate makes sense because:
(1.) they're two separate actions
(2.) existing users may have fully installed but still not be sure how to update

It is not the exact same as an install, especially on the recommended Wally platform.

Maybe we can review the future of the updating article separately. My goal with this patch is to simplify the docs layout, because it's currently too complicated to be usable.

> it also causes the documentation to feel incredibly unorganized and unstructured.

There isn't really a logical structure that works here. Showing the user everything all at once means they can easily pick exactly what they want. "Commands" is an article users will reference when working, so it doesn't make sense to be consigned to a step-by-step 'setup' guide, even if that is a purpose it serves.

Other articles (like security) do exist only for reference, but they might still be important to some beginner users. The goal with the flattened structure is to stop trying to graft on a folder hierarchy and let users select what articles they need. We still gently guide users, my choice of ordering was fairly intentional, but I'm not convinced there is a way to meet user needs with categorisation.

Personally this kind of flattened structure is not super unfamiliar to me. Especially given how, all in all, few articles we have. *KISS.*

> why is updating just in the middle of the list rather than at the top?

The idea (you'll notice this if you look at the file name numbers) is that updating sits at the end of the step-by-step guide. Updating is at the end on purpose, it's something you do periodically, but 


--
Hope that all makes sense.



On Mon, 24 Aug 2026, at 23:55, Wil wrote:
> wilyt1 left a comment (evaera/Cmdr#432) <https://github.com/evaera/Cmdr/pull/432#issuecomment-5402595814>
> What if we merge the contents of the Updating page into installation?

ShadowDaughter:
I originally wrote a huge wall of text going over each one of your points, but in all honesty, reading this all through just makes me realize that the current documentation has already been quite unorganized even before this change and it should heavily be reconsidered how we explain and document everything. Here are some noteworthy points:
1. Why is there a page explaining what types are, which built-in types exist, their structure and how to create them, just for the proper in-depth explanation of their creation to be explained in another page further down in an "advanced" section?
2. Commands, Meta-Commands and AutoExec are all very related and yet are thrown across the documentation in different sections, with little explanation on how they actually correlate and the importance of using them together.
3. Hooks, while listed pretty early, are for some reason still only shown after the guide on commands, despite being the most important security aspect and needing to probably be mentioned as soon as possible. At the same time, guards, which are essentially just mini-hooks for commands, are explained way further down in an "advanced" section, even tho they are just as relevant as hooks and important for the usage of commands.
4. Where do we differenciate between something belonging into the cookbook or being worth of a own page? Even more complicated things such as custom type annotations are thrown into the cookbook, but changes that seem like a rather simple thing such as customizing the interface have a own page dedicated to what are just a few sentences.
5. Where do we use a step-by-step guide and where do we simply explain features independantly? The page on Commands and Types are more-so feature explanations that explain the general usage and functionality of these, while however being included in the the section for the step-by-step guide.

While these issues aren't necessarily relevant to this issue/PR directly, I feel like they should still be noted since they play a huge role in why the documentation feels so awkward to use. While I can see that removing categories and making the entire documentation flat can make quick access easier, I believe the real issue is the way the documentation is structured instead.

In my rant I mentioned that the pages on commands as example feel all over the place. Why don't we have a category dedicated to specifically commands? If I'm not mistaken, it is possible to have a category itself act as a page with information (which could contain the main commands page) and a container for further pages, where we could include meta-commands and autoexec.

autonordev:
These sorts of points are exactly what I had in mind. But fixing them would be so onerous. We make do with what we have.

As you point out, user journeys are unclear (e.g. guards being 'advanced', because they're outside of the step-by-step, even though they're just... a normal feature) and articles serve both as tutorials and references (_Commands_, for example).

A docs overhaul would be great. But it would require a lot of planning, starting with the question of _what do people use the Cmdr docs for_. And while I'm happy to support, I don't think I have the bandwidth to do that.

The goal with this patch is to make the best of the current situation. I think it does that.

wilyt1 (approved):
The flattened layout looks good to me. We can revisit grouping related concepts post-1.13.

PR #435, opened 2026-09-03
Fix `:exit` re-enabling chat UI the game had disabled

sqikerz:
Running `:exit` (or any `Cmdr:Hide()`) before the window has ever been opened force-enabled `ChatWindowConfiguration`, `ChatInputBarConfiguration`, and `ChannelTabsConfiguration`, unhiding chat UI in games that deliberately disable it. 

`Window:_setChatEnabled(false)` is what saves the previous state, so if the window was never shown, the saved values are `nil` and the restore branch fell back to `true`.

**Declarations**:


wilyt1 (approved):

PR #440, opened 2026-09-09
fix: HasCharacter guard rejecting streamed-out players

sqikerz:
With `Workspace.StreamingEnabled`, a distant player's character isn't replicated to the client, so the `HasCharacter` guard read "streamed out" as "no character" and killed `teleport` before it reached the server. The client now defers on players it can't see and lets the server decide, and `position` falls back to the server as well as `teleport` only counts players it actually moved.

**Declarations**:

### DIFF
--- a/Cmdr/BuiltInCommands/Admin/teleportServer.luau
+++ b/Cmdr/BuiltInCommands/Admin/teleportServer.luau
@@ -1,12 +1,15 @@
 const SEAT_TIMEOUT = 1
 
 const function teleportCharacter(humanoid: Humanoid, character: Model, cFrame: CFrame): boolean
-	if humanoid.SeatPart then
+	if humanoid.Sit then
+		const wasSeated = humanoid.SeatPart ~= nil
 		humanoid.Sit = false
 
-		const deadline = os.clock() + SEAT_TIMEOUT
-		while humanoid.SeatPart and os.clock() < deadline do
-			task.wait()
+		if wasSeated then
+			const deadline = os.clock() + SEAT_TIMEOUT
+			while humanoid.SeatPart and os.clock() < deadline do
+				task.wait()
+			end
 		end
 	end
 
--- a/Cmdr/BuiltInCommands/Admin/teleportServer.luau
+++ b/Cmdr/BuiltInCommands/Admin/teleportServer.luau
@@ -8,7 +8,7 @@ return function(_, fromPlayers: { Player }, destination: Player | Vector3): stri
 	for _, player: Player in ipairs(fromPlayers) do
 		const character = player.Character
 		const humanoid = character and character:FindFirstChildWhichIsA("Humanoid") :: Humanoid?
-		if not character or not humanoid then
+		if not humanoid then
 			continue
 		end
 
--- a/Cmdr/BuiltInCommands/Admin/teleportServer.luau
+++ b/Cmdr/BuiltInCommands/Admin/teleportServer.luau
@@ -3,12 +3,16 @@ return function(_, fromPlayers: { Player }, destination: Player | Vector3): stri
 		then (destination.Character :: Model):GetPivot()
 		else CFrame.new(destination)
 
+	local teleported = 0
+
 	for _, player: Player in ipairs(fromPlayers) do
 		const character = player.Character
-		const humanoid = character and character:FindFirstChildWhichIsA("Humanoid")
-		if not humanoid then
+		const humanoid = character and character:FindFirstChildWhichIsA("Humanoid") :: Humanoid?
+		if not character or not humanoid then
 			continue
 		end
+
+		teleported += 1
 
 		task.spawn(function()
 			if humanoid.SeatPart then
@@ -22,5 +26,9 @@ return function(_, fromPlayers: { Player }, destination: Player | Vector3): stri
 		end)
 	end
 
-	return `Teleported {#fromPlayers} players.`
+	if teleported == #fromPlayers then
+		return `Teleported {teleported} players.`
+	end
+
+	return `Teleported {teleported} of {#fromPlayers} players. The rest had no character.`
 end
--- a/Cmdr/BuiltInCommands/Admin/teleportServer.luau
+++ b/Cmdr/BuiltInCommands/Admin/teleportServer.luau
@@ -1,8 +1,29 @@
+const SEAT_TIMEOUT = 1
+
+const function teleportCharacter(humanoid: Humanoid, character: Model, cFrame: CFrame): boolean
+	if humanoid.SeatPart then
+		humanoid.Sit = false
+
+		const deadline = os.clock() + SEAT_TIMEOUT
+		while humanoid.SeatPart and os.clock() < deadline do
+			task.wait()
+		end
+	end
+
+	if not character:IsDescendantOf(workspace) then
+		return false
+	end
+
+	character:PivotTo(cFrame)
+	return true
+end
+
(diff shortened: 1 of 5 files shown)


Cmdr/CmdrClient/Shared/BuiltInGuards/HasCharacter.luau
```diff
@@ -19,9 +24,11 @@ return function(index: number?)
...
 
 		const players: { Player } = if isSingle then { target :: Player } else target :: { Player }
+		const canVerifyEveryone = IS_SERVER or not workspace.StreamingEnabled
 
 		for _, player in ipairs(players) do
-			if hasValidCharacter(player) then
+			const canVerify = canVerifyEveryone or player == Players.LocalPlayer
+			if not canVerify or hasValidCharacter(player) then
 				continue
 			end
```
ShadowDaughter, line 33: Shouldn't we instead simply prevent this guard from running on the client at all? The guard should realistically only ever run on the server, and we can save the client from running code that the server will run again later anyways.

Cmdr/BuiltInCommands/Admin/teleportServer.luau
```diff
@@ -3,13 +3,17 @@ return function(_, fromPlayers: { Player }, destination: Player | Vector3): stri
...
 		else CFrame.new(destination)
 
+	local teleported = 0
+
 	for _, player: Player in ipairs(fromPlayers) do
 		const character = player.Character
-		const humanoid = character and character:FindFirstChildWhichIsA("Humanoid")
-		if not humanoid then
+		const humanoid = character and character:FindFirstChildWhichIsA("Humanoid") :: Humanoid?
+		if not character or not humanoid then
```
ShadowDaughter: Checking for the character is sort of unnecessary here, the humanoid cannot exist if the character doesn't.
```suggestion
		if not humanoid then
```

Cmdr/BuiltInCommands/Debug/position.luau
```diff
@@ -21,10 +21,17 @@ return {
...
+		if rootPart then
+			return (tostring(rootPart.Position):gsub("%s", ""))
+		end
 
-		return tostring(rootPart.Position):gsub("%s", "")
+		if player == Players.LocalPlayer then
+			return "You don't have a character."
+		end
+
+		return nil
```
ShadowDaughter: Should we consider omitting the ClientRun portion of this command entirely? While it in theory reduces the strain on the server, the code being performed on the server is negligible and would unify the implementation into one location.

ShadowDaughter (commented):
LGTM overall, but I wanted to leave some comments anyways.

Cmdr/CmdrClient/Shared/BuiltInGuards/HasCharacter.luau
sqikerz: That'll break ClientRun-only commands such as blink / thru, deference to the server is handled correctly in the current state I believe, where the client is allowed to be the source of truth only in scenarios where it can be (ie; streaming disabled OR checking self). This also is much faster than waiting the roundtrip for the server for no reason, if we were to have client commands such as blink/thru go there.

Cmdr/BuiltInCommands/Admin/teleportServer.luau
```diff
@@ -3,13 +3,17 @@ return function(_, fromPlayers: { Player }, destination: Player | Vector3): stri
...
+
 	for _, player: Player in ipairs(fromPlayers) do
 		const character = player.Character
-		const humanoid = character and character:FindFirstChildWhichIsA("Humanoid")
+		const humanoid = character and character:FindFirstChildWhichIsA("Humanoid") :: Humanoid?
 		if not humanoid then
 			continue
 		end
 
+		teleported += 1
```
wilyt1: If the character leaves `workspace` while waiting for `SeatPart` to clear, `character:PivotTo(cFrame)` is skipped, but the player has already been counted.

wilyt1 (commented):
See comment below.

Cmdr/BuiltInCommands/Admin/teleportServer.luau
```diff
@@ -1,26 +1,58 @@
+const SEAT_TIMEOUT = 1
+
+const function teleportCharacter(humanoid: Humanoid, character: Model, cFrame: CFrame): boolean
+	if humanoid.SeatPart then
+		humanoid.Sit = false
+
+		const deadline = os.clock() + SEAT_TIMEOUT
+		while humanoid.SeatPart and os.clock() < deadline do
```
autonordev: Are there instances in which case `Sit = false` is called but no `SeatPart` property change signal is sent?

Cmdr/BuiltInCommands/Debug/position.luau
```diff
@@ -20,11 +20,4 @@ return {
 	Guards = {
 		HasCharacter(1),
 	},
-
-	ClientRun = function(_, player: Player): string
```
autonordev, line 24: Would it not be preferable for us to keep this code as is, and then _if_ character is nil _and_ a workspace uses `StreamingEnabled` hand off to the server to do the response? This would lead to faster responses whenever possible, which is esp. important bcs `pos` is used in metaprogramming

Cmdr/BuiltInCommands/Admin/teleportServer.luau
sqikerz: Yes, the property being manually set (such as a `sit` command).

docs/51-guards.md
```diff
@@ -52,6 +52,8 @@ Cmdr processes guards sequentially in the order they are listed inside the `Guar
 
 Because command definitions are shared between client and server, guards are evaluated on the client first during UI invocation, and evaluated **again on the server** prior to running server code to prevent exploiters from bypassing client checks.
 
+Keep in mind that the client doesn't see everything the server does. With [`Workspace.StreamingEnabled`](https://create.roblox.com/docs/reference/engine/classes/Workspace#StreamingEnabled), for example, another player's character may not be replicated to the executor's client, so a client-side check can't tell the difference between "not there" and "not visible to me". Guards which inspect state the client might be missing should defer to the server (return `nil`) instead of rejecting the command, as the server check runs regardless.
```
autonordev:
```suggestion
Keep in mind that the client doesn't see everything the server does. For example, with [`Workspace.StreamingEnabled`](https://create.roblox.com/docs/reference/engine/classes/Workspace#StreamingEnabled) another player's character might not be replicated to the executor's client, so a client-side check can't tell the difference between "not there" and "not visible to me". Guards which inspect state the client might be missing should defer to the server (return `nil`) instead of rejecting the command, as the server check runs regardless.
```

autonordev (changes requested):
Hey @sqikerz, thanks for contributing. A few structural comments, I think these are bits that need to be improved, and a wording nit.

Cmdr/BuiltInCommands/Admin/teleportServer.luau
sqikerz: I'll have it be event based in that thread off the property as that seems like the better angle.

sqikerz: Actually there's no need to hang for something that doesn't block the `PivotTo()`, and an event here would just be a waste.

wilyt1: The event is necessary so that `PivotTo()` doesn't move the seat with the character.

wilyt1 (approved):

PR #443, opened 2026-09-13
Character in input fix (Window.luau)

Versus41:
When CMDR toggle is set to an imputable character (like ";"), said character will appear in input window. This fork fixes the issue by adding a 1 frame wait time before enabling the window.

**Declarations**:

- [X] I declare that this contribution was created in whole or in part by me.


Versus41:
closing cause other pull had better fix

PR #445, opened 2026-09-21
Allow for RegisterCommandsIn to have a Custom script filter (This can allow you to change how it finds which scripts to use)

Cosmella-v:
Adds a new argument for RegisterCommandsIn to filter which scripts would count as "Server" and which Scripts count as "Client"
This would allow you register scripts using the name differently
One filter example is having Client.luau and having that link to Server.luau
```luau
const myCustomFilter = function(context)
	const commandScript = context.Script
	if commandScript then
		const scriptName = commandScript.Name
		if not scriptName:find("Server") then
			if scriptName == "Client" then
				return commandScript, context.Container:FindFirstChild("Server")
			end
			return commandScript, context.Container:FindFirstChild(`{commandScript.Name}Server`)
		else
			return nil, commandScript
		end
	else
		return nil, nil
	end
end
Cmdr.Registry:RegisterCommandsIn(PathToMyCommands, nil, nil, myCustomFilter)
```

**Declarations**:


ShadowDaughter:
.. But why tho? Stating that "you like it this way" is no real reason. The code you provided is also not on par with Cmdr's expected quality.

Cosmella-v:
I cant give image examples right now since im on my phone but

- When registering a command it searches the folder(s) but you couldn't really setup something like that without drawbacks
- Also i believe it feels more intuative to have it like this

Folder name
-> Command name folder
-> -> Server.luau
-> -> Client.luau
Instead you would have to do something to work around the folder "based" method 
-> Command name folder
-> -> Command.luau
-> -> CommandServer.luau

Which i think cmdr should support Client.luau being able to link to Server.luau Instead of ClientServer.luau

Also I think i can fix the code quality when I get home, i thought of a solution that might be nicer.
(It may take awhile tho)

autonordev:
Can you meaningfully explain why this would be a significant improvement to Cmdr users?

“I like it this way” is not a good enough reason, I **don’t** like it this way! I think this is a bad idea, and devs who use it will end up adding another vscode alias to show the folder name (as you do with init.luau)

FWIW you can already do this anyway, just with a structure like Example -> Example.luau, ExampleServer.luau (I have used this structure myself).

ShadowDaughter:
Another thing to note (as per one of the staff in the Discord) is that you _force_ this to be the new default behavior, which is an unwelcome change most developers won't enjoy.

Cosmella-v:
> Can you meaningfully explain why this would be a significant improvement to Cmdr users?
> 
> “I like it this way” is not a good enough reason, I **don’t** like it this way! I think this is a bad idea, and devs who use it will end up adding another vscode alias to show the folder name (as you do with init.luau)
> 
> FWIW you can already do this anyway, just with a structure like Example -> Example.luau, ExampleServer.luau (I have used this structure myself).

I have redone this pull request to allow you to have a custom filter of modules
This allows you have to have better customization within without replacing the default behaviour (by default)

This change would allow someone to change how module loading works if they **"don't like how it works"**

An example with the template code
image

sqikerz:
> Another thing to note (as per one of the staff in the Discord) is that you _force_ this to be the new default behavior, which is an unwelcome change most developers won't enjoy.

Yeah, this would have to bump into a new major version as it's a breaking change for everyone, and also this doesn't seem very optimal for a breaking change..

ShadowDaughter:
@Cosmella-v Even with your "improvements" I don't see the real purpose behind this change. The `RegisterCommandsIn` function is a quick way for developers to register commands without further hassle. Adding extra layers of complexity to this (such as through a custom filter) takes away its point, and you can already achieve what you're doing here by simply writing your own method for registering commands that relies on the other methods of the Registry API.

